PR #1, opened 2020-07-23
Added support to build against specific R versions

jbabyhacker:
Utilizes R_HOME environment variable to determine R version to build against.


andy-thomason:
Thanks for the help. We'll be reliant on others to check different environments.

PR #11, opened 2020-11-22
enable easy extraction of created bindings

clauswilke:
I have patched the build script to also save the bindings to a folder specified by `$LIBRSYS_BINDINGS_DIR`, if it is set. This makes it easy to generate and cache the bindings.


andy-thomason (approved):
Looks good.

PR #12, opened 2020-11-28
use R_HOME if available

clauswilke:
When `R_HOME` is set we should just use it rather than try to run R to find it. In fact, when R builds extension packages it explicitly disables running R without a complete path.


jbabyhacker:
How did you determine the standard path layout?

PR #15, opened 2020-11-29
Clippy and rustc-cfg improvements

jbabyhacker:
* Added ignore rules so clippy doesn't complain.
* Added rustc-cfg to pass the binding version to extendr to support building against R 3.x versions.
* Bumped version

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ matrix:
 
 addons:
   apt:
-    packages: # recommanded versions for rust-bindgen
+    packages: # recommended versions for rust-bindgen
       - r-base-dev
       - llvm-3.9-dev
       - libclang-3.9-dev
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -1,6 +1,6 @@
 [package]
 name = "libR-sys"
-version = "0.1.10"
+version = "0.1.11"
 authors = ["andy-thomason <andy@andythomason.com>"]
 edition = "2018"
 description = "Low level bindings to the R programming language."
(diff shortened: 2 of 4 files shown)


clauswilke:
This seems reasonable in principle, but could you maybe first open an issue explaining the problem and the proposed solution?

Also, when you open a PR, please keep updating it rather than closing it and opening a new one. (I'm talking about #14, which looks to be the same as this one.)

clauswilke:
Also, as a general principle, it's best to keep separate issues in separate PRs. Are the clippy directives related to the new feature you're adding or are they unrelated? If they are unrelated they belong into a separate PR, with its own related issue explaining the problem.

jbabyhacker:
The problem is that the bindings generated are different between R versions. So if you use symbols in extendr that are not supported in the underlying R version you build against, then extendr will fail to build.

The clippy issues are unrelated to the changes I added. I just discovered them when modifying the code.

jbabyhacker:
I will be putting up a change into extendr that allows for extendr to work on older 3.x versions of R.

clauswilke:
That's all fine and good, but then the first step should be to open an issue for extendr about the specific problem you're concerned about. Issues are for problems, PRs are for solutions. It's a good idea to always start with the problem, because there may be multiple solutions to the same problem, or sometimes a PR seems to solve a problem but later it reappears.

andy-thomason:
This branch may be a bit stale.

This is still a good idea, so we should get all projects to pass clippy and format requirements.

clauswilke:
Yes, I'm going to close. This PR is too outdated relative to where the codebase now is.

@jbabyhacker If you'd like to address the clippy issues, please open a new PR just about that.

PR #17, opened 2020-12-09
set up GH actions

clauswilke:
set up GH actions


clauswilke:
GitHub actions build and test successfully on all three OSs (Linux, macOS, Windows).

Issue #16, opened 2020-12-08
 Path determination fails on Windows


clauswilke:
If `R_HOME` isn't set, we try to find the path to R by executing these lines:
https://github.com/extendr/libR-sys/blob/887b9aab97b6f5163c358193641fca4661fc992c/build.rs#L54

However, this doesn't work, because Windows mangles the paths. See here:
https://github.com/extendr/extendr/issues/67#issuecomment-741190771

Thanks to @Ilia-Kosenkov for discovering this. If we can find a way for R to print a valid Windows path we can fix this problem. Otherwise, `R_HOME` needs to be set for Windows builds to succeed.

lebensterben:
https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-getfullpathnamea

lebensterben:
`R.home` calls an internal `R.home` method:
https://github.com/wch/r-source/blob/bcd5629813f4e82c965a07280288423f3a758362/src/library/base/R/files.R#L19-L36

which calls the C function `R_Home`:
https://github.com/wch/r-source/blob/bcd5629813f4e82c965a07280288423f3a758362/src/main/names.c#L799

that in term calls `R_HomeDir`:
https://github.com/wch/r-source/blob/bcd5629813f4e82c965a07280288423f3a758362/src/main/platform.c#L1462-L1469

well, eventually it check the environment variable `R_HOME`
https://github.com/wch/r-source/blob/bcd5629813f4e82c965a07280288423f3a758362/src/main/sysutils.c#L225-L228

I'm not sure why calling `R.home` will have different result from checking `R_HOME` directly.

clauswilke:
This appears to have been resolved with #18.

PR #18, opened 2020-12-09
Improved path setup

clauswilke:
The reworked code only tries to find the top-level R folder and then assumes a standard directory structure underneath. This should fix some problems on Windows.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -84,10 +84,6 @@ fn probe_r_paths() -> io::Result<InstallationPaths> {
                     .to_string()
             };
 
-            println!("R home: {}", r_home);
-            println!("R library: {}", library);
-            println!("R binary: {}", r_binary);
-
             let out = Command::new(&r_binary)
                 .args(&[
                     "-s",
--- a/build.rs
+++ b/build.rs
@@ -92,10 +92,11 @@ fn probe_r_paths() -> io::Result<InstallationPaths> {
                 .args(&[
                     "-s",
                     "-e",
-                    r#"cat(normalizePath(R.home("include")), sep = '\n')"#
+                    r#"cat(normalizePath(R.home('include')), sep = '\n')"#
                 ])
                 .output()?;
 
+            // if there are any errors we print them out, helps with debugging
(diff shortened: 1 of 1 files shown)


clauswilke:
Well this was harder than I thought. On Linux the includes can live in a different location entirely, and we need to be prepared for that.

@yutannihilation Could you take a look? I think it's all good, but maybe there's something I've overlooked.

yutannihilation:
Sure, I think I can have some time to take a look tomorrow. I setup my old Windows machine again :)

One thing I'm a bit afraid is stdout is not encoded in UTF-8, which means the following line is not correct, but I'm not sure...

``` rust
             let rout = String::from_utf8_lossy(&rout.stdout); 
```

andy-thomason:
If you have shift-jis on Windows (or even ANSI) this could be a problem. We
should use
the OSPath strings for filenames and correct path building.

clauswilke:
> We should use the OSPath strings for filenames and correct path building.

I'm happy to look into that. However, since this has implications for pretty much every part of the build script, I'd suggest to handle it in a separate PR. The current PR fixes the logic of finding the paths. The next PR can fix the problem of strings that are not valid UTF-8.

andy-thomason:
Windows file path handling was a huge problem we faced at Sony
as windows file functions used ANSI instead of UTF8.

Behaviour would vary if you were running on Japanese Windows or En-US
Windows!

clauswilke:
I've filed an issue specifically about this: #19

yutannihilation (approved):
Looks good to me! I confirmed this works on my Windows laptop.

Also thanks for filing a new issue.

clauswilke:
@yutannihilation Thanks! Can you confirm that it works without setting `R_HOME`? Can I close #16?

yutannihilation:
> Can you confirm that it works without setting `R_HOME`? Can I close #16?

Yes, I meant, I confirmed `cargo build` and `cargo test` succeed on Command Prompt without `R_HOME` set. I believe we can close #16 :tada:

PR #22, opened 2020-12-11
Use PathBuf and OsString to represent OS native strings in build script

### LINKED ISSUE #19: Properly deal with non-UTF-8 strings in build script
The current build script will likely not succeed on certain locales on Windows. For context, see here:
https://github.com/extendr/libR-sys/pull/18#issuecomment-741873561

clauswilke:
An attempt to fix #19 by replacing `String` with `OsString` in the build script.

It is not successful yet. Two problems I have encountered:

- There is no simple way to convert `&Vec[u8]` into `OsString`. Some suggestions on how this can be done are available here: https://doc.rust-lang.org/stable/std/ffi/index.html#conversions It is OS specific.

- There is no simple way to print out an `OsString`, as it doesn't implement the `Format` trait. We'll probably have to write our own printing function. This may again be OS specific.

@andy-thomason @yutannihilation Any input, advice, suggestions would be appreciated.


lebensterben:
you should use Path instead of OsStr or OsString

First, Path can be cheaply created from OsStr because Path is also Os-independent.

https://doc.rust-lang.org/std/path/struct.Path.html#method.new

Then,

the output should be valid utf-8, so you can use String::from_utf8 to get String, and then use <String as AsRef<Path>>::as_ref

clauswilke:
Thanks, this addresses the second point. The first point still needs some work on Windows.

lebensterben:
Output of a Command is always Vec<u8> so you can convert is to a String, which can be converted to Path/PathBuf.

If you explicitly need an OsStr/OsString, that can be cheaply get from Path.

I don't understand what's the "first point" that's not solved

clauswilke:
The first point is converting `Vec<u8>` to `OsString` without assuming that `Vec<u8>` holds valid UTF-8 and without doing lossy conversion.

lebensterben:
Okay, to be clear. You can replace String::from_utf_lossy by String::from_utf

There's no assumption made.

clauswilke:
How do you know the Ok value is always valid utf-8?

From the documentation (https://doc.rust-lang.org/std/process/struct.Output.html#fields):
> `stdout: Vec<u8>`
> The data that the process wrote to stdout.

The documentation makes no guarantee regarding valid utf-8. Presumably, if the process writes something else that's what we get.

lebensterben:
Command::output use std::io::Read::read to get the result from a anonymous pipe.

If it behaves correctly, it will promptly return an Error when the result is not valid utf-8, including the case that the child process is terminated but the output doesn't terminate properly.

clauswilke:
> Command::output use std::io::Read::read to get the result from a anonymous pipe.
> 
> If it behaves correctly, it will promptly return an Error when the result is not valid utf-8, including the case that the child process is terminated but the output doesn't terminate properly.

Please provide evidence. The documentation doesn't say anything about such a guarantee: https://doc.rust-lang.org/std/io/trait.Read.html#tymethod.read It also doesn't make logical sense. Why would the `read()` function operate on byte buffers if it guarantees utf-8? That's what the `String` type is for.

clauswilke:
This works now.

@yutannihilation Are you in a position to test this with file names that aren't encoded as utf-8?

yutannihilation:
Thanks, I'm happy to test this. But, unfortunately (or fortunately?) my path to `R.exe` and `R_HOME` contain only ASCII characters, so I might need some time to setup CJK path to see if this works or not.

On GitHub Actions Windows runner, I think it's possible to write some tests using paths containing latin characters. This might complicates the CI scrpits, though.
c.f. https://github.com/tidyverse/dplyr/blob/07d5a7fc1155b00afa611b8e6c6562c80223ab53/tests/testthat/helper-encoding.R#L3

clauswilke:
Would it be possible to just copy everything under `R_HOME` into a folder with non-ASCII characters and then set `R_HOME` to that folder before building?

Alternatively, though, if the code looks good to you we could also merge and see if anybody complains.

yutannihilation:
I thought I can just use a symlink, but then `normalizePath()` returns the original path so it doesn't make sense for testing. So I ended up renaming my `R_HOME`. Anyway, I confirmed this pull request fixes the problem 🎉 

### My setup

``` console
>R -s -e "cat(normalizePath(R.home()))"
C:\Program Files\R\錆
```

### Current master

``` console
>set PATH=C:\msys64\usr\bin;C:\msys64\mingw64\bin;%PATH%
>cargo test
   Compiling libR-sys v0.1.10 (C:\Users\hiroaki-yutani\Documents\repo\R\libR-sys)
error: failed to run custom build command for `libR-sys v0.1.10 (C:\Users\hiroaki-yutani\Documents\repo\R\libR-sys)`

Caused by:
  process didn't exit successfully: `C:\Users\hiroaki-yutani\Documents\repo\R\libR-sys\target\debug\build\libR-sys-9d5d2beaccf19f42\build-script-build` (exit code: 1)
  --- stdout
  Problem locating local R install: Os { code: 2, kind: NotFound, message: "指定されたファイルが見つかりません。" }
```

### This pull request

``` console
>set PATH=C:\msys64\usr\bin;C:\msys64\mingw64\bin;%PATH%
>cargo test
    Updating crates.io index
   Compiling libR-sys v0.1.10 (C:\Users\hiroaki-yutani\Documents\repo\R\libR-sys)
    Finished test [unoptimized + debuginfo] target(s) in 17.49s
     Running target\debug\deps\libR_sys-25defe17c96e6128.exe

running 19 tests
test bindgen_test_layout_R_CMethodDef ... ok
test bindgen_test_layout_R_CallMethodDef ... ok
test bindgen_test_layout_R_inpstream_st ... ok
test bindgen_test_layout_R_outpstream_st ... ok
test bindgen_test_layout_Rcomplex ... ok
test bindgen_test_layout___mingw_dbl_type_t ... ok
test bindgen_test_layout___mingw_dbl_type_t__bindgen_ty_1 ... ok
test bindgen_test_layout___mingw_flt_type_t ... ok
test bindgen_test_layout___mingw_ldbl_type_t ... ok
test bindgen_test_layout___mingw_ldbl_type_t__bindgen_ty_1 ... ok
test bindgen_test_layout__complex ... ok
test bindgen_test_layout__exception ... ok
test bindgen_test_layout__iobuf ... ok
test bindgen_test_layout_localeinfo_struct ... ok
test bindgen_test_layout_max_align_t ... ok
test bindgen_test_layout_tagLC_ID ... ok
test bindgen_test_layout_threadlocaleinfostruct ... ok
test bindgen_test_layout_threadlocaleinfostruct__bindgen_ty_1 ... ok
[1] 1
test tests::test_eval ... ok

test result: ok. 19 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out

   Doc-tests libR-sys

running 1 test
test src\lib.rs - (line 18) ... ok

test result: ok. 1 passed; 0 failed; 0 ignored; 0 measured; 0 filtered out
```

build.rs
```diff
@@ -23,85 +73,72 @@ fn probe_r_paths() -> io::Result<InstallationPaths> {
...
-                    r#"cat(normalizePath(R.home('include')), sep = '\n')"#
+                    r#"cat(normalizePath(R.home('include')))"#
                 ])
                 .output()?;
 
             // if there are any errors we print them out, helps with debugging
-            for errln in String::from_utf8_lossy(&out.stderr).lines() {
-                println!("> {}", errln);
+            if !out.stderr.is_empty() {
+                println!("> {}", String::from_utf8_lossy(&out.stderr));
```
yutannihilation, line 134: Can we use `byte_array_to_os_string()` here as well?

yutannihilation (approved):
I have little knowledge how to handle OS-dependent encodings properly in Rust, but this looks good to me as far as I can tell.

build.rs
clauswilke: I don't think so. (Or at least I don't know how.) `OsString` doesn't implement the `Display` trait, so I can't print it. In the rest of the code I got around this issue by using `PathBuf`, which does implement `Display`. But we can't convert an arbitrary message into a `PathBuf`. I really don't understand why `OsString` doesn't implement `Display`.

yutannihilation: I meant something like `byte_array_to_os_string(&out.stderr).as_os_str().to_string_lossy()`. `String::from_utf8_lossy(&out.stderr)` is trying to interpret the non-UTF-8 byte array as UTF-8, which seems wrong, if I understand correctly.

clauswilke: Ah, I see. Yes, it's possible that this is somewhat more reliable. I'll make the change.

PR #23, opened 2020-12-12
WIP: trying to turn on 32-bit compilation on Windows in GH Actions

clauswilke:
This is an experiment. Not immediately expected to work.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -72,10 +72,6 @@ jobs:
           update: false
           install: >-
             mingw-w64-x86_64-clang
-            #mingw-w64-x86_64-headers-git
-            #mingw-w64-x86_64-make
-            #mingw-w64-i686-gcc
-            #mingw-w64-i686-headers-git
       - uses: r-lib/actions/setup-r@v1
         with:
           r-version: 'release'
@@ -115,8 +111,6 @@ jobs:
           update: false
           install: >-
             mingw-w64-i686-clang
-            #mingw-w64-i686-gcc
-            #mingw-w64-i686-headers-git
       - uses: r-lib/actions/setup-r@v1
         with:
           r-version: 'release'
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -70,7 +70,6 @@ jobs:
       - uses: actions-rs/toolchain@v1
         with:
           toolchain: ${{ matrix.rust-version }}
-          #target: x86_64-pc-windows-gnu
           default: true
           components: rustfmt, clippy
       - name: Build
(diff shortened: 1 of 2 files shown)


clauswilke:
@CGMossa @yutannihilation @Ilia-Kosenkov I have managed to compile libR-sys for i686 on Windows. Strangely it only works if we provide clang with a special include directory; otherwise it fails:
https://github.com/extendr/libR-sys/blob/4ef44e70bef61df431442f106814ea93bb3bbb5d/build.rs#L163-L167
You'd think that if the build target is set right it can figure this out correctly, but apparently it can't.

If any of you can think of a way of getting this to work without this hack I'd be interested to know. Otherwise I'll make this configurable via an environment variable.

CGMossa:
Yes this is the reoccurring problem. For me, this is where stdio is located. I don't know how cross compilation is supposed to work.. But running 
mingw32 should help. Right now, you're running mingw64 as per 
```
msystem: MINGW64
```

.github/workflows/test.yml
```diff
@@ -75,17 +85,30 @@ jobs:
...
+          cargo build -vv --target i686-pc-windows-gnu
         env:
           LIBCLANG_PATH: C:/msys64/mingw64/bin
       - name: Run tests
         run: |
           echo "C:/msys64/usr/bin" >> $GITHUB_PATH
           echo "C:/msys64/mingw64/bin" >> $GITHUB_PATH
-          cargo test -- --nocapture --test-threads=1
+          cargo test --target x86_64-pc-windows-gnu -- --nocapture --test-threads=1
+          echo "C:/msys64/mingw32/bin" >> $GITHUB_PATH
```
yutannihilation, line 110: [GitHub Actions doc](https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-commands-for-github-actions#adding-a-system-path) says:

> Prepends a directory to the system `PATH` variable for all subsequent actions in the current job. The currently running action cannot access the new path variable.

So, I think this doesn't work. Maybe we can just set `PATH` envvar by `$ENV:Path="...;"+$ENV:Path`? (Is the shell PowerShell here, right...?)

clauswilke: Ah, maybe I just have to set it once during the setup. I'll try it.

clauswilke: So that helped, actually. Now the test compiles but then crashes. :-) It also means all these `echo` statements were not needed.

clauswilke: The error is 0xc000007b `INVALID_IMAGE_FORMAT`, which is apparently a 32-bit/64-bit mismatch: https://social.msdn.microsoft.com/Forums/vstudio/en-US/c1147429-95f8-47e7-a532-f3960118ce11/x86-application-gives-invalidimageformat-0xc000007b-error-on-windows-7-64-bit?forum=clr

So something still isn't quite right with the setup.

clauswilke:
For some reason actions have stopped running for this PR but still run in my personal branch. I'm also now getting the "unable to find libclang" error (https://github.com/clauswilke/libR-sys/runs/1543272073?check_suite_focus=true#step:6:1083):
```
[libR-sys 0.1.10] thread 'main' panicked at 'Unable to find libclang: "the `libclang` shared library at C:\\msys64\\mingw32\\bin\\libclang.dll could not be opened: The specified procedure could not be found. (os error 127)"', C:\Users\runneradmin\.cargo\registry\src\github.com-1285ae84e5963aae\bindgen-0.53.3\src/lib.rs:1956:31
[libR-sys 0.1.10] note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
error: failed to run custom build command for `libR-sys v0.1.10 (D:\a\libR-sys\libR-sys)`
```
This error means the build script compiles but then doesn't run properly. The problem is that the shared library can't be opened. I don't think this is a path error. It finds the library but there's a mismatch. Maybe it compiled against a different version than it now tries to link?

In the mean time, I discovered a few things that have moved me further along. Most importantly, we can set the target in the toolchain name, and that resolves a lot of issues with incorrect default targets: 
https://github.com/extendr/libR-sys/blob/9a947a7c00a796715665fc630a627e118ebba21e/.github/workflows/test.yml#L94
It also resolves the issue of missing `stdio.h` when compiling the build script.

I'm going to clean up with github action for x86_64, and then I may go back to trying to get this to work for i686. In either case, I think for now it's best to have two separate action runners for 64 bit and 32 bit. If we can't get it to work in two separate runners we'll certainly not get it to work in one.

clauswilke:
Retiring this PR in favor of #25, which continues where this one left of.

PR #27, opened 2020-12-14
Provide precomputed bindings

### LINKED ISSUE #26: Infrastructure for precomputed bindings
Given the various issues we have with compilation on Windows, I think we need tools for precomputed bindings as soon as possible. So I've started implementing this. See here: https://github.com/clauswilke/libR-sys/tree/precomputed_bindings

Generate bindings by setting the `LIBRSYS_BINDINGS_OUTPUT_PATH` variable to provide an output location:
```
LIBRSYS_BINDINGS_OUTPUT_PATH=~/Desktop cargo build
```

Then move the bindings into the `bindings` folder of the crate: https://github.com/clauswilke/libR-sys/tree/precomputed_bindings/bindings

Then build and test with these bindings by setting the `--no-default-features` flag (you may have to run `cargo clean` to trigger the rebuild):
```
cargo build --no-default-features
cargo test --no-default-features
```

I'd be particularly interested in knowing whether it's possible to generate bindings under 64-bit windows and use under 32-bit windows. If somebody could test that that would be great. I assume you'll have to change the architecture name from `x86_64` to `x86` in the bindings file name for this to work.

cc @Ilia-Kosenkov @CGMossa @yutannihilation

clauswilke:
This closes #26. See that issue for details.

@andy-thomason As this is a near-complete rewrite of the build script, I have added myself to the authors and bumped the version number to 0.2.0. Please indicate whether you approve.


andy-thomason (approved):
Nice one, @clauswilke .

If the precompiled bindings work, we can also lose a few dependencies.

I guess this is necessary for windows.

By all means add yourself to the copyright holders.

PR #32, opened 2020-12-18
Updating Windows description in README

Ilia-Kosenkov:
Restructures section dedicated to building for  `Windows`.
Includes both `CMD` and `PowerShell` commands (where applicable).

Also, utilizes a feature available in `cargo` if installed via `rustup`:
`cargo +toolchain build --target=target` allows to build using `toolchain` without setting default toolchain, which is very important if we want to easily build for 32 and 64 bits.

### DIFF
--- a/README.md
+++ b/README.md
@@ -21,7 +21,7 @@ Two components are required to build the library:
 2. [rust](https://www.rust-lang.org/learn/get-started): It is recommended to install `rust` using `rustup`; search path should include `rust` binaries.
 
 
-When building for `Windows`, the default host shall be `stable-msvc` and special `rust` targets should be added for compatibility with `R`:
+When building for `Windows`, the default host should be `stable-msvc` and special `rust` targets should be added for compatibility with `R`:
 - **Windows**
   ```Shell
   rustup default stable-msvc
--- a/README.md
+++ b/README.md
@@ -21,9 +21,10 @@ Two components are required to build the library:
 2. [rust](https://www.rust-lang.org/learn/get-started): It is recommended to install `rust` using `rustup`; search path should include `rust` binaries.
 
 
-When building for `Windows`, special `rust` targets should be added for compatibility with `R`:
+When building for `Windows`, the default host shall be `stable-msvc` and special `rust` targets should be added for compatibility with `R`:
 - **Windows**
   ```Shell
+  rustup default stable-msvc
   rustup target add x86_64-pc-windows-gnu  # 64-bit
   rustup target add i686-pc-windows-gnu    # 32-bit
   ```
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ To test the build, run `cargo test`. Note the `--test-threads=1`, without this f
     cargo test -- --nocapture --test-threads=1
     ```
 - **Windows**
-    **NOTE:** currently `Windows` tests are unstable, but still can be executed with additional effort:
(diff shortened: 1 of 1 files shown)


clauswilke:
This is very useful. Could you make a few more changes though? This includes insight I've generated from working on this PR: https://github.com/extendr/extendr/pull/78 (You may want to review the GH Action script to see what I'm doing).

1. I now believe things only work reliably on Windows when the default host is a msvc toolchain. We may want to recommend this.
2. The installation is much simpler on all platforms when people use the precomputed binding. So that should be the primary documentation, in a separate section. Install R; Install Rust; build; on all platforms. No need to install msys2.
3. The information about using bindgen should go into a separate section.

Thanks!

Ilia-Kosenkov:
@clauswilke,
I checked your GH Action script.
If I understood you correctly, I can write a unified installation section that utilizes simple and recommended precomputed bindings and then move what is currently written to the `bindgen` section (with modifications where necessary). If this seems reasonable, I will do this

clauswilke:
Yes, that's exactly what I mean. Thanks!

Ilia-Kosenkov:
@clauswilke,
I have updated README. 
I believe we need someone with experience with/access to MacOs/Linux to update respective sections, as right now readme is Windows-dominated :)

README.md
```diff
@@ -9,76 +9,247 @@ Low-level R library bindings
...
+    </details>
+
+
+## Building bindings from source (advanced)
+
+The bindings can be generated using [bindgen](https://github.com/rust-lang/rust-bindgen), special `rust` crate. 
+`bindgen` usage is enabled via `use-bindgen` feature flag.
+
+`bindgen` requires [libclang](https://clang.llvm.org/docs/Tooling.html), which should be installed first. 
+This library relies on `LIBCALNG_PATH` environment variable to determine path to the appropriate version of `libclang`.
```
clauswilke, line 100: Please note the typo here: `LIBCLANG_PATH`

clauswilke (commented):
Looks good. Notice the one typo I flagged. And maybe state at the beginning that for windows, the default host should be `pc-windows-msvc`, not `pc-windows-gnu`.

README.md
Ilia-Kosenkov: Fixed in the followup commits

README.md
```diff
@@ -9,76 +9,248 @@ Low-level R library bindings
...
-1. [R](https://cran.r-project.org/): It needs to be installed and available in the search path.
-2. [libclang](https://clang.llvm.org/docs/Tooling.html): Depending on your operating system, you may need to set the `LIBCLANG_PATH` environment variable or add `llvm-config` to your search path.
+## Using precomputed bindings (recommended)
 
+Two components are required to build the library:
+1. [R](https://cran.r-project.org/): It needs to be installed and available in the search path.
+2. [rust](https://www.rust-lang.org/learn/get-started): It is recommended to install `rust` using `rustup`; search path should include `rust` binaries.
+
+
+When building for `Windows`, the default host shall be `stable-msvc` and special `rust` targets should be added for compatibility with `R`:
```
clauswilke, line 24: Sorry, one last nitpick: "shall" sounds very old-fashioned and biblical. It's probably better to either write "should" or, if you want to make a stronger statement, "must".

Ilia-Kosenkov: Done

clauswilke (approved):
Thanks!

Issue #29, opened 2020-12-16
Calling R from Rust on Windows 32bit is broken


clauswilke:
The latest version of the libR-sys precompiled bindings for Windows compile on 32 bit (https://github.com/extendr/helloextendr/runs/1561836399?check_suite_focus=true#step:11:174), but `cargo test` still fails for libR-sys (https://github.com/extendr/libR-sys/runs/1561658815#step:7:218), with an error that suggests a linking problem. I'm starting to think that this is because the code we use to call R from Rust is a bit naive, and basically assumes a unix-like system:
https://github.com/extendr/libR-sys/blob/73e57498c238615072f0742347599517a29726d7/src/lib.rs#L74-L86
(I'm linking here to code where I've tried to fix this up a bit, but still there may be a fundamental problem with this approach on Windows, in particular on 32 bit.)

The R documentation describes two entirely different approaches to embedding R from external code, one for Unix and one for Windows:
https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Linking-GUIs-and-other-front_002dends-to-R
 
Also, on systems with multiple architectures (such as Windows) we should not assume that the call to `Rf_initialize_R()` can automatically find the correct runtime to start:
https://cran.r-project.org/doc/manuals/R-admin.html#Sub_002darchitectures
Setting the `--arch` flag may help (but didn't fix the issues here).

clauswilke:
@Ilia-Kosenkov Would you mind reproducing this workflow:
https://github.com/clauswilke/libR-sys/blob/73e57498c238615072f0742347599517a29726d7/.github/workflows/test.yml#L139
using this branch:
https://github.com/clauswilke/libR-sys/tree/precomputed_bindings
and testing what happens that causes this crash:
https://github.com/extendr/libR-sys/pull/27/checks?check_run_id=1561658815#step:7:218

It looks like a library mismatch or missing symbol in one of the libraries, so stepping through the loading process like you did the other day should uncover what the problem is. I suspect that either the 32-bit Rust binary is trying to link to the 64-bit R runtime or that it is trying to call a function that simply doesn't exist in 32-bit R.

clauswilke:
The branch I linked to in the previous comment doesn't exist anymore. It's been merged into https://github.com/extendr/libR-sys/.

Ilia-Kosenkov:
@clauswilke,
I somehow missed this issue, I will try reproducing if it still makes sense.

clauswilke:
[...]

Issue #9, opened 2020-11-19
Building on i686-pc-windows-gnu


CGMossa:
It is important that we can build on 32-bit targets, I've tried to follow the instructions in the `README.md`,
but substituting with the 32-bit variants.

This doesn't work for some reason. And here's the latest error-message/hurdle:

```
[redacted] MINGW32 /c/Users/tpb398/Documents/GitHub/libR-sys
$ cargo build --target=i686-pc-windows-gnu
   Compiling libR-sys v0.1.9 (C:\Users\tpb398\Documents\GitHub\libR-sys)
error: failed to run custom build command for `libR-sys v0.1.9 (C:\Users\tpb398\Documents\GitHub\libR-sys)`

Caused by:
  process didn't exit successfully: `C:/cargo_target_dir\debug\build\libR-sys-46304186790a078b\build-script-build` (exit code: 101)
  --- stdout
  cargo:rustc-env=R_HOME=C:/R/R-4.0.2
  cargo:rustc-env=R_HOME=C:/R/R-4.0.2/bin/i386
  cargo:r_home=C:/R/R-4.0.2
  cargo:rustc-link-search=C:/R/R-4.0.2/bin/i386
  cargo:rustc-link-lib=dylib=R
  cargo:rerun-if-changed=build.rs
  cargo:rerun-if-changed=wrapper.h

  --- stderr
  thread 'main' panicked at 'Unable to find libclang: "couldn\'t find any valid shared libraries matching: [\'clang.dll\', \'libclang.dll\'], set the `LIBCLANG_PATH` environment variable to a path where one of these files can be found (invalid: [(C:\\Users\\tpb398\\scoop\\apps\\msys2\\current\\mingw32\\bin\\libclang.dll: invalid DLL (32-bit)), (C:\\Users\\tpb398\\scoop\\apps\\msys2\\current\\mingw32\\bin\\libclang.dll: invalid DLL (32-bit))])"', C:\Users\tpb398\.cargo\registry\src\github.com-1ecc6299db9ec823\bindgen-0.53.3\src/lib.rs:1956:31
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

I've searched for `clang.dll` in the mingw folders, and there is nothing named that. 

It is the `libclang.dll` that is problem.

Here's a few more diagnostic things that I see mentioned when trying to troubleshoot this:
<details>

```bash
$ clang -v
[...]

PR #34, opened 2020-12-21
Update Github Actions for Windows

Ilia-Kosenkov:
Using insights from #9 we can now more or less reliably generate bindings for both 32- and 64-bit Windows.
With the path trick discussed in #29 `i386/i686` tests can be run (though most of the tests are disabled).

I suggest to split Windows job into two separate ones and build bindings separately for each architecture.
I am not sure if I correctly copied the binding upload process, this should be checked.


clauswilke:
Thanks for your efforts on this one. I have two comments:

1\. I'm not a big fan of breaking the windows builds into two separate runs because most of their time is spent setting up the R and Rust environments, so there's a lot of wasted compute. I know we get the GH Action hours for free but we should still be cognizant of resource use.

2\. The action scripts need to be completely reworked along the lines of the extendr scripts so we can target specific R versions:
https://github.com/extendr/extendr/blob/master/.github/workflows/test.yml
We will eventually need at least `release`, `devel`, and `oldrel` for all architectures, I think. See e.g. here:
https://github.com/r-lib/actions/blob/master/examples/README.md#tidyverse-ci-workflow

It may make sense to put everything into one single matrix or to have two matrices, one for unix-like builds and one for windows builds. Not sure about that one, depends on how much special casing is needed.

If you'd like to take a stab at this, you're welcome to do so.

clauswilke:
To be clear: It's the following part of the extendr test script that would make a good starting point for libR-sys:
https://github.com/extendr/extendr/blob/d27585f43f900c49f7a805be6ce3cd83a196ab6d/.github/workflows/test.yml#L112-L149
We don't need to install and cache R packages.

Ilia-Kosenkov:
@clauswilke ,
I was thinking about merging everything into a single job, but decided to progress in small steps.

I can try to rewrite script following `extendr`as an example. If I fail or someone else does this faster, we can safely discard this PR.

Ilia-Kosenkov:
So this is a template for a new GH Actions script, inspired by `extendr`.
There are two things to address: first, `ubuntu::r-devel` fails tests, which I presume also happens for other repos and is not our fault; second -- I am not sure how bindings and artifacts are now generated -- before we had one job for each os, now -- multiple. Perhaps there should be some condition on which to generate bindings -- for instance, only for `rust-stable`.

clauswilke:
To make tests work on r-devel you need to set `LD_LIBRARY_PATH`, see here:
https://github.com/extendr/extendr/blob/06bd2a400894eb8fe9a43360afb6da1a1c8d7a30/.github/workflows/test.yml#L155-L156
(This is most likely the reason why you'll have to treat Unix and Windows separately.)

For the artifacts, there's nothing to worry about, generating and saving the artifacts costs practically nothing. I think it's good to have artifacts for all settings we test. This allows us to check whether anything changes, e.g., whether bindings under `nightly` look different from those under `stable`.

.github/workflows/test.yml
```diff
@@ -11,152 +11,119 @@ on:
...
-        env:
-          LIBRSYS_BINDINGS_OUTPUT_PATH: generated_bindings
-      - name: Run tests
-        run: cargo test -vv --features use-bindgen -- --nocapture --test-threads=1
-      - name: Upload generated bindings shipped with libR-sys
-        if: steps.build.outcome == 'success'
-        uses: actions/upload-artifact@main
+          r-version: ${{ matrix.config.r }}
+
+      - name: Set up `rust`
```
clauswilke, line 49: I think this should be `Rust` instead of `` `rust` ``. In general, I think names of programming languages should not be set in backticks, but more specifically here I don't think the backticks are converted into code font.

.github/workflows/test.yml
```diff
@@ -11,152 +11,119 @@ on:
...
-      run:
-        shell: msys2 {0}
-    strategy:
-      matrix:
-        rust-version: [stable-x86_64-pc-windows-gnu]
-    steps:
-      - uses: actions/checkout@v2
-      - uses: msys2/setup-msys2@v2
+      # This step is only needed for Windows
+      - name: Set up `msys2` for Windows
```
clauswilke, line 57: I would similarly delete backticks here.

.github/workflows/test.yml
```diff
@@ -11,152 +11,119 @@ on:
...
-#          default: true
-#          components: rustfmt, clippy
-#      - name: Build with precomputed bindings
-#        run: cargo build -vv
-#      - name: Run tests on precomputed bindings
-#        run: cargo test -vv -- --nocapture --test-threads=1
 
+      # Run tests again using different bindings
+      - name: Run tests on precomputed bindings shipped with libR-sys
+        run: cargo test -vv $env:CARGO_BUILD_FLAGS -- --nocapture --test-threads=1
```
clauswilke, line 129: Please add a newline at the end of the file.

Ilia-Kosenkov:
I noticed I screwed up with `matrix.config`. I will check everything is working correctly and fix all issues.

Ilia-Kosenkov:
Debugging GitHub actions is a real pain.
@clauswilke, I think I fixed everything you mentioned, take a look when you have time.

.github/workflows/test.yml
```diff
@@ -11,152 +11,127 @@ on:
...
       matrix:
-        rust-version: [stable, beta, nightly]
+        config:
+          - {os: windows-latest, r: 'release', rust-version: 'stable-x86_64-pc-windows-msvc'}
+          - {os: windows-latest, r: 'release', rust-version: 'stable-i686-pc-windows-msvc'}
+          - {os: macOS-latest,   r: 'release', rust-version: 'stable'}
+          - {os: ubuntu-20.04,   r: 'release', rust-version: 'stable', rspm: "https://packagemanager.rstudio.com/cran/__linux__/focal/latest"}
+          - {os: ubuntu-20.04,   r: 'release', rust-version: 'nightly', rspm: "https://packagemanager.rstudio.com/cran/__linux__/focal/latest"}
+          # R-devel requires LD_LIBRARY_PATH
+          - {os: ubuntu-20.04,   r: 'devel',   rust-version: 'nightly', rspm: "https://packagemanager.rstudio.com/cran/__linux__/focal/latest"}
```
clauswilke, line 30: I think R devel should be tested with rust stable, so if something breaks we know whether it's R devel or rust nightly that is the problem. Also, could you expand the matrix so there is an R `oldrel` run for each OS/architecture? I'd like to have bindings for those versions as well (I realize this will break the test of precompiled bindings; don't worry about that or download the artifacts and add them, as you wish).

.github/workflows/test.yml
```diff
@@ -11,152 +11,127 @@ on:
...
+
+
+      # If success, upload bindings
       - name: Upload generated bindings
-        if: steps.build.outcome == 'success'
+        if: steps.build.outcome == 'success' && steps.test.outcome == 'success' && startsWith(matrix.config.rust-version, 'stable')
         uses: actions/upload-artifact@main
         with:
-          name: Windows generated bindings
+          name: ${{ matrix.config.os }} ${{ matrix.config.rust-version }} generated bindings
```
clauswilke, line 132: The generated bindings need to have the R version in the name also.

clauswilke (commented):
Overall looks good. Just a few remaining items while you're at it.

Don't worry about squashing the commits, GitHub does that automatically when merging.

.github/workflows/test.yml
Ilia-Kosenkov: Ok, we now test three versions of R for each architecture. I also added R-release rust-nightly generating no bindings to test rust features.

Actually, you were wrong about tests -- you (or someone else) implemented a fallback to default bindings if the exact version is missing, so tests were successful (and a warning was logged). I downloaded new bindings, now we have fallback, 4.0 release, 4.1 devel, and 3.6 oldrel versions for each target.

.github/workflows/test.yml
Ilia-Kosenkov: Artifact names are now identical to job names

clauswilke:
You're correct. I added the fallback if bindings are missing. Thanks!

PR #36, opened 2020-12-22
Contributing guidelines and code of conduct

clauswilke:
I have created contributing guidelines and code of conduct. I envision that we could use these throughout the extendr project.

@andy-thomason @yutannihilation Could you review and let me know if you see any issues or have suggestions for additional items that should be mentioned?

### DIFF
--- a/CODE-OF-CONDUCT.md
+++ b/CODE-OF-CONDUCT.md
@@ -1,0 +1,134 @@
+
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+We as members, contributors, and leaders pledge to make participation in our
+community a harassment-free experience for everyone, regardless of age, body
+size, visible or invisible disability, ethnicity, sex characteristics, gender
+identity and expression, level of experience, education, socio-economic status,
+nationality, personal appearance, race, religion, or sexual identity
+and orientation.
+
+We pledge to act and interact in ways that contribute to an open, welcoming,
+diverse, inclusive, and healthy community.
+
+## Our Standards
+
+Examples of behavior that contributes to a positive environment for our
+community include:
+
+* Demonstrating empathy and kindness toward other people
+* Being respectful of differing opinions, viewpoints, and experiences
+* Giving and gracefully accepting constructive feedback
+* Accepting responsibility and apologizing to those affected by our mistakes,
+  and learning from the experience
+* Focusing on what is best not just for us as individuals, but for the
+  overall community
+
+Examples of unacceptable behavior include:
+
+* The use of sexualized language or imagery, and sexual attention or
+  advances of any kind
+* Trolling, insulting or derogatory comments, and personal or political attacks
+* Public or private harassment
+* Publishing others' private information, such as a physical or email
+  address, without their explicit permission
+* Other conduct which could reasonably be considered inappropriate in a
+  professional setting
+
+## Enforcement Responsibilities
+
+Community leaders are responsible for clarifying and enforcing our standards of
+acceptable behavior and will take appropriate and fair corrective action in
+response to any behavior that they deem inappropriate, threatening, offensive,
+or harmful.
+
+Community leaders have the right and responsibility to remove, edit, or reject
+comments, commits, code, wiki edits, issues, and other contributions that are
+not aligned to this Code of Conduct, and will communicate reasons for moderation
+decisions when appropriate.
+
+## Scope
+
+This Code of Conduct applies within all community spaces, and also applies when
+an individual is officially representing the community in public spaces.
+Examples of representing our community include using an official e-mail address,
+posting via an official social media account, or acting as an appointed
+representative at an online or offline event.
+
+## Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be
+reported to the community leaders responsible for enforcement at
+[INSERT CONTACT METHOD].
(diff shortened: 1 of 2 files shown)


CODE-OF-CONDUCT.md
```diff
@@ -0,0 +1,132 @@
+
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+We as members, contributors, and leaders pledge to make participation in our
```
yutannihilation, line 6: What are "members", "contributors", and "leaders"?

CODE-OF-CONDUCT.md
```diff
@@ -0,0 +1,132 @@
...
+* Trolling, insulting or derogatory comments, and personal or political attacks
+* Public or private harassment
+* Publishing others' private information, such as a physical or email
+  address, without their explicit permission
+* Other conduct which could reasonably be considered inappropriate in a
+  professional setting
+
+## Enforcement Responsibilities
+
+Community leaders are responsible for clarifying and enforcing our standards of
```
yutannihilation, line 42: Which should we use "community" or "project"? I feel both can fit here, but have different nuances.

CODE-OF-CONDUCT.md
```diff
@@ -0,0 +1,132 @@
...
+This Code of Conduct applies within all community spaces, and also applies when
+an individual is officially representing the community in public spaces.
+Examples of representing our community include using an official e-mail address,
+posting via an official social media account, or acting as an appointed
+representative at an online or offline event.
+
+## Enforcement
+
+Instances of abusive, harassing, or otherwise unacceptable behavior may be
+reported to the current maintainers of the project via email (see `Cargo.toml` file for Rust crates, `DESCRIPTION` file for R packages). All complaints will be reviewed and investigated promptly and fairly.
```
yutannihilation, line 63: As the "Authorship" section of "CONTRIBUTING.md" says, authors are those who "have made multiple, sustained, and/or non-trivial contributions to the project." Just being listed on authors field doesn't mean that they have the will and bandwidth to maintain the project.

yutannihilation (commented):
Thanks, basically looks good to me. I'm not sure I'm the right person to review this (I made only trivial contributions so far), but let me leave some comments mainly for discussion. The terminologies might be a bit ambiguous when it's applied to this project and I think we can improve them.

> I envision that we could use these throughout the extendr project.

What about creating a separate repo (e.g. "meta"?) and place the file there? The individual repo's CoC can just link to them so that we don't need to fix the same typo (or more serious changes) here and there.

CODE-OF-CONDUCT.md
clauswilke: The original document I adapted this from envisioned providing a contact email here. We can always just list @andy-thomason's email. :-) More seriously, though, I think it's always best to have multiple contact points. I'm open to suggestions on how to set this up better.

As an example, the ggplot2 code of conduct uses the following wording:
> Instances of abusive, harassing, or otherwise unacceptable behavior may be reported by opening an issue, emailing the project lead (hadley@rstudio.com), or one of the core developers.

CODE-OF-CONDUCT.md
clauswilke: Maybe at this point we have a project rather than a community?

CODE-OF-CONDUCT.md
clauswilke: The ggplot2 Code of Conduct, which used a much earlier version of this document, writes:
> As contributors and maintainers of this project, ...

Maybe that's better?

clauswilke:
I have considered moving these documents into a separate repo, but I think that causes important usability issues that outweigh the added maintenance cost. If these documents are plain text, then they travel with the repo and are immediately available wherever somebody has a copy of the repo. If they are links, at a minimum they require an extra click, and in the worst case scenario the link may be broken or the network down. I'd want to make these documents as immediately available as possible, in particular the contributing guide.

In addition, over time these documents may reasonable diverge between different repos, e.g. different people may take the lead maintaining different repos or different repos may adopt slightly different styles of operation.

CODE-OF-CONDUCT.md
clauswilke: I notice that crates.io has the concept of "owners" in addition to "authors". Maybe we say something like "see Owners on [crates.io](https://crates.io/crates/libR-sys)".

yutannihilation:
> I think that causes important usability issues that outweigh the added maintenance cost.

I see. I agree with this point.

> over time these documents may reasonable diverge between different repos

I think `CODE-OF-CONDUCT.md` shouldn't diverge. As the scope of it isn't within the single repository, it should provide consistent rules. `CONTRIBUTING.md` is fine.

CODE-OF-CONDUCT.md
yutannihilation: A project sounds better (at least at the moment) to me.

CODE-OF-CONDUCT.md
yutannihilation: I think so, it sounds clearer to me.

clauswilke:
@yutannihilation I believe I have addressed all your comments. Could you read it over one more time?

yutannihilation:
Thanks, I confirmed all my comments are addressed!

andy-thomason (approved):

PR #40, opened 2020-12-22
Fixing i686 tests

### LINKED ISSUE #39: Enabling Win-i686 tests
Currently, bindgen layout tests are disabled for Windows x86
https://github.com/extendr/libR-sys/blob/80488ec76147ae367de8d8050107998771dcfb3f/build.rs#L260-L263

There are two failing layout tests out of 19 actually, and I found similar issues, for instance, https://github.com/rust-lang/rust-bindgen/issues/1823. This issue was not resolved, because the problem itself is severe: https://github.com/rust-lang/rust/issues/54341, if I understand correctly, `LLVM` produces incorrect (or different) alignment for `int128`/`float128` types.

We can enable the majority of tests if we blacklist `max_align_t` (mentioned in `rust-lang` issue) and `__mingw_ldbl_type_t`, which, I assume, is `long double`, which is misaligned in `LLVM`. 
This also means that any usage of these types is unsafe as they are not ABI compatible with `C` counterparts, but I am not sure how widely they are used.

Ilia-Kosenkov:
Addresses, but does not fully fix issue #39.
Enables bindgen layout test but blacklists two problematic types (for Win-x86) only.

New precomputed bindings with tests for Win-x86.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -3,7 +3,8 @@ name = "libR-sys"
 version = "0.1.12"
 authors = [
     "andy-thomason <andy@andythomason.com>",
-    "Claus O. Wilke <wilke@austin.utexas.edu>"
+    "Claus O. Wilke <wilke@austin.utexas.edu>",
+    "Ilia Kosenkov <ilia.kosenkov.at.gm@gmail.com>"
 ]
 edition = "2018"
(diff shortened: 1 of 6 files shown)


clauswilke (approved):
Thanks!

PR #42, opened 2020-12-25
Removing unnecessary MSYS2 step from GHA

### LINKED ISSUE #41: [Windows] MSYS2 appears to be part of the image
If I understand [this listing](https://github.com/actions/virtual-environments/blob/main/images/win/Windows2019-Readme.md) correctly, `MSYS2` is bundled with the `windows-latest` image, but it is not added to the `PATH`.
Since we configure `MSYS2` paths ourselves, it appears we do not need `MSYS2` installation step.
Assuming it is not affected by caching, it works fine without it, see 
https://github.com/Ilia-Kosenkov/libR-sys/runs/1606864716

This may save us about 10-12 seconds, which is not much, but if I am correct, we just waste resources each time.
Again, if this is true, other build pipelines (at least, `extendr`) are also affected.

Ilia-Kosenkov:
Closes #41.
Renamed all steps in my fork to bypass possible caching to confirm it works.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -71,26 +71,15 @@ jobs:
           default: true
           components: rustfmt, clippy
 
-      # This step is only needed for Windows
-      # - name: Set up MSYS2 for Windows
-      #   if: runner.os == 'Windows'
-      #   uses: msys2/setup-msys2@v2
-      #   with:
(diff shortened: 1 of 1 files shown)


clauswilke:
Thanks!

PR #45, opened 2020-12-29
Cross-compiling and precomputed bindings for Mac-aarch64 

### LINKED ISSUE #35: libR-sys bindings for Apple M1
Sooner rather than later, we will need bindings for Apple M1 chips. One strategy we could pursue is to cross-compile, as we're already doing for our Windows bindings. The only downside is that we can't test whether the bindings actually work.

Here's a relevant issue about Github Action runners on M1: https://github.com/actions/virtual-environments/issues/2187

Ilia-Kosenkov:
Partially fixes 35.
`aarch64` is only supported by `rust-nightly`, so a separate VM with MacOS-11.0, R-release and rust-nightly is used to generate bindings.


clauswilke:
Could you also add a version of the bindings with R version removed from the name? The generic bindings are simply a copy of the most recent version-specific bindings. (Ideally I'd want to set this up as a symbolic link but those don't work well in git repos.)

clauswilke (approved):

clauswilke:
Thanks! I approve, so you can go ahead and merge.

Ilia-Kosenkov:
I'll wait for macos-11 test first before proceeding, thanks.

PR #49, opened 2021-01-02
Output R version info

### LINKED ISSUE #33: Export R version to downstream crates
Currently we sort-of export the R version to downstream crates, but only when using bindgen:
https://github.com/extendr/libR-sys/blob/6ff5c6c496ed38143599bd1801b2c7999c765173/build.rs#L275

I think we should always export the R version, and probably also in the form major/minor/patch. However, this brings up the question with precomputed bindings of whether we export the R version that bindgen was run with or the R version that the current compilation is using. I am leaning towards the latter, as I think it'll generally be more useful. It allows down-grading of bindings: Let's say I have precomputed bindings for R 3.6 and I'm using R 3.5 and there's a function in 3.6 that is not in 3.5. Then conditionally not compiling anything that would use this function should work. The fact that the function is declared in the bindings should not matter.

In the reverse situation, where we report the version of the bindings rather than of the current R, we could up-grade bindings, but I think it's much less likely to be useful. We could never use a feature that is not in the bindings anyways, so conditional compile doesn't help.

clauswilke:
Closes #33.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## libR-sys devel
 
+- Output R version info to downstream crates using variables `R_VERSION_MAJOR`,
+ `R_VERSION_MINOR`, `R_VERSION_PATCH`, `R_VERSION_DEVEL`, and `R_VERSION_STRING`.
+
 - Added precomputed bindings for Apple Silicon.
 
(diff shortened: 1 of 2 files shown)


Ilia-Kosenkov (approved):
Everything seems good to me!

Issue #21, opened 2020-12-10
[Windows] Is it possible to include Rust toolchain in Rtools?


yutannihilation:
Currently libR-sys requires users to install standalone MSYS2 for libclang, but as https://github.com/extendr/extendr/issues/13 suggests, it might be good to use Rtools' MSYS2 for simplicity.

Rtools' MSYS uses a dedicated repository, which has only the limited number of packages available. To add some packages there, we can send a pull request to [r-windows/rtools-packages repo](https://github.com/r-windows/rtools-packages), according to this document: [System Libraries in Rtools40](https://github.com/r-windows/docs/blob/master/rtools40.md).

clauswilke:
Reading through the documents for Rtools, I wonder whether we may have to start a conversation with the maintainers first. 

In [this section](https://github.com/r-windows/rtools-packages#contributing-to-rtools-packages) it says that we can only contribute static libraries, but that probably won't work for libclang.

In [this section](https://github.com/r-windows/rtools-packages#extra-msys-build-utilities) it says that we can alternatively provide additional build tools but those should only be used to build more packman packages. But we need additional build tools to be used during the R package build.

@yutannihilation Do you want to initiate this conversation with the Rtools maintainers?

yutannihilation:
> In [this section](https://github.com/r-windows/rtools-packages#contributing-to-rtools-packages) it says that we can only contribute static libraries, but that probably won't work for libclang.

Thanks, I didn't notice this... After some thinking, I now feel this is less important.

* For Windows user who just use CRAN packages, they can install CRAN precompiled binaries, so this won't affect them.
* For Windows user who try non-CRAN pakcages, they might be able to use a prebuilt static library (which seems undesirable according to https://github.com/extendr/extendr/pull/10#issuecomment-724577056). If we don't choose this approach, we might want to use Rtools' MSYS2 for the ease of installation.
* For Windows developers, probably they need to install Rust toolchain anyway for debugging etc.

Let's hold this issue on hold for a while.

clauswilke:
Yes, I think we're good at this time. It might be more important to add Rust than libclang.

I've also been thinking that long-term, if this project gets widely used, we should try to convince the R core team to distribute the bindings as part of their official distribution. In this way we'd know that we always have the correct bindings for each R version and platform.

Ilia-Kosenkov:
Rtools cannot be used for `libclang`. However, it can be used with precomputed bindings, see discussion in
[...]

PR #51, opened 2021-01-17
Using rtools to test precomputed bindings on Windows

Ilia-Kosenkov:
After discussing this issue in https://github.com/extendr/rextendr/issues/19, a solution was found to build & test `libR-sys` using only rtools. This does not work for bindings generation (no clang provided with rtools), but is suitable for user consumption.

Currently, rtools mingw64 should be "fixed" (see https://github.com/r-windows/rtools-packages/blob/master/mingw-w64-gcc/PKGBUILD#L305-L309) by duplicating `libgcc.a` as `libgcc_eh.a`, which satisfies `rustc` and linker.

Build & test were reordered. 
First, rtools is configured and precomputed bindings are tested.
Second, msys2 is enabled and bindings generation is tested using msys2.
This does not affect other systems.

Partially addresses #21.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -130,6 +130,26 @@ jobs:
           Del alias:R
           echo "LD_LIBRARY_PATH=$(R -s -e 'cat(normalizePath(R.home()))')/lib" | Out-File -FilePath $env:GITHUB_ENV -Encoding utf8 -Append
 
+      # Run tests again using different bindings
+      - name: Run tests on precomputed bindings shipped with libR-sys
+        run: |
+          foreach ($target in ($env:RUST_TARGETS).Split(",")) {
+            if(($env:NO_TEST_TARGETS).Split(",").Contains($target)) {
+              echo "::warning:: Skipping tests for target: $target"
+            }
+            else {            
+              echo "::group::Running tests for target: $target"
+              cargo test -vv $(if ($target -ne 'default') {"--target=$target"} ) -- --nocapture --test-threads=1
+              echo "::endgroup::"
+              if (!$?) { 
+                echo "::error::Running tests for target: $target";
+                throw "Last exit code $LASTEXITCODE"
+              }
+            }
+          }
+        env: 
+          NO_TEST_TARGETS: ${{ join(matrix.config.no-test-targets, ',') }}
+
       # Build and emit bindings to './generated_bindings'
       - name: Build & Emit bindings
         id: build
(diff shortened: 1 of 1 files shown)


clauswilke:
I think it's good to know that this works, but I'm not sure I would merge this at this time. We should just reach out to the Rtools maintainers and ask them to include this fix.

I wouldn't want to tell our endusers that they need to manually copy library files (it will look scary to them) and I think our GH Actions should mirror the installation protocol we recommend in our manuals.

Ilia-Kosenkov:
@clauswilke,
Ok but I suggest to leave it open for reference purposes for now.

clauswilke:
I'd also be Ok with adding one runner that does it this way, but without deleting the msys2-based runners. This way we can keep an eye on both options until Rtools is fixed.

.github/workflows/test.yml
```diff
@@ -135,10 +135,11 @@ jobs:
         id: build
         run: |
           foreach ($target in ($env:RUST_TARGETS).Split(",")) {
-            echo "Building for target: $target"
+            echo "::group::Building for target: $target"
             cargo build -vv --features use-bindgen $(if ($target -ne 'default') {"--target=$target"} )
+            echo "::endgroup::"
```
clauswilke, line 140: Doesn't the endgroup statement have to come after the `if` statement checking for errors?

Ilia-Kosenkov: I put "groups" around `cargo` commands so log under spoiler contains only build/test related output.
If `if` condition is met and error is thrown, it will be visible separtely.

Ilia-Kosenkov:
I added a copy of the workflow with a reduced number of steps to test rtools build as a separate job, restored the original workflow and added `"::group::"` statements to improve logging of exceptionally long verbose `cargo` outputs.

.github/workflows/test.yml
clauswilke: Yes, but see: https://github.com/extendr/extendr/pull/130#discussion_r558802290
Does this not apply here?

Ilia-Kosenkov: Oh, I see, that is indeed very stupid of me. I will fix it with the next commit.

clauswilke (approved):

.github/workflows/test.yml
```diff
@@ -190,13 +192,105 @@ jobs:
...
+        env:
+          RUST_TARGETS: ${{ join(matrix.config.targets, ',')}}
+      
+      # All configurations for Windows go here
+      # Rust toolchain is used to determine target architecture
+      - name: Configure Windows
+        if: runner.os == 'Windows'
+        # 1. Configure path to libclang
+        # 2. Add path to mingw32/mingw64 -- otherwise library is linked to rtools
+        # 3. Add path to R's i386/x64  -- to solve x86 build/test issue
```
clauswilke, line 262: Sorry, missed this earlier: The comments look like they need updating.

.github/workflows/test.yml
```diff
@@ -190,13 +192,105 @@ jobs:
...
+            echo "C:\rtools40\mingw64\bin" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append ;
+            echo "$(Rscript.exe -e 'cat(normalizePath(R.home()))')\bin\x64"  | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append ;
+          }
+          if ($env:RUST_TARGETS -like "*i686*") {
+            echo "C:\rtools40\mingw32\bin" | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append ;
+            echo "$(Rscript.exe -e 'cat(normalizePath(R.home()))')\bin\i386"  | Out-File -FilePath $env:GITHUB_PATH -Encoding utf8 -Append ;
+          }
+          
+
+      # Run tests again using different bindings
```
clauswilke, line 276: Same here: comment looks wrong.

PR #52, opened 2021-01-17
Remove kernel32-sys

yutannihilation:
Just happened to find this. winapi's README says:

> Should I still use those -sys crates such as kernel32-sys?
> 
> No. Those crates are a legacy of how winapi 0.2 was organized. Starting with winapi 0.3 all definitions are directly in winapi itself, and so there is no longer any need to use those -sys crates.

So I guess we only need to use winapi, right? (I didn't even know winapi is a successor of kernel32-sys until today...)


clauswilke (approved):
Great! fewer dependencies should mean faster builds.

PR #56, opened 2021-05-05
Using `Rtools40v2` to test bindings on Windows

### LINKED ISSUE #53: [Windows] [CI] Use Rtools to test pre-computed bindings
This is a reminder to update CI workflow
https://github.com/extendr/libR-sys/blob/ccdf5d57071ee058b3ed5d7cd62054204e11c722/.github/workflows/test.yml#L264-L266
in light of the recent merging of this PR https://github.com/r-windows/rtools-packages/pull/186.

The problem on Windows side is that MSYS2 (used for generating bindings), Rtools (custom MSYS2 build) and mingw toolchains in rust's `*pc-windows-gnu` ship different versions of gcc and friends. This creates issues that can be resolved using hacks (see code snippet above).
In particular, in PR https://github.com/r-windows/rtools-packages/pull/186 we suggested implementing a hack that was present in 32-bit version of Rtools, but not in 64-bit version.
As soon as updated rtools gets released (I believe it will take significant time), we can update CI here. I hope with Rtools patched we may be able to move additional workflows to Rtools only, avoiding both MSYS2 and rust's own `mingw` tools.

Ilia-Kosenkov:
This fixes #53.
[Rtools40v2](https://cran.r-project.org/bin/windows/Rtools/) has been recently released.
It includes our [hack](https://github.com/r-windows/rtools-packages/pull/186) to please rust compiler and resolves `gcc` linking issue.

`libR-sys` can now be safely tested using Rtools only (no dependence on `msys2` is required for pre-computed bindings).

However, the `setup-r` [action](https://github.com/r-lib/actions) currently have a bug that does not allow installing other versions of Rtools40 even if the version string is explicitly specified (see my issue here https://github.com/r-lib/actions/issues/289).
There is also a PR to make the new version default (https://github.com/r-lib/actions/pull/287), which will not work until the issue is resolved.

I suggest to leave this PR as a draft for now. As soon as `setup-r` gets updated, we will remove the relevant code (marked with comments in the `.yaml`) and will finally have a working setup:
https://github.com/Ilia-Kosenkov/libR-sys/blob/a55515f4a999adfa4535aaed57112e3a233864ef/.github/workflows/test.yml#L215-L227

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -213,7 +213,10 @@ jobs:
     steps:
       
       - uses: actions/checkout@v2
-      
+      # https://github.com/r-lib/actions/pull/287
+      - name: Purge system Rtools installation
+        run: rm C:\rtools40 -Force -Recurse
+
       - name: Set up R
         uses: r-lib/actions/setup-r@v1
         with:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -211,8 +211,10 @@ jobs:
     steps:
       
       - uses: actions/checkout@v2
+
       # When this issue is resolved and PR merged, this step can be safely removed
       # https://github.com/r-lib/actions/pull/287
+      # https://github.com/r-lib/actions/issues/289
       - name: Purge system Rtools installation
         run: rm C:\rtools40 -Force -Recurse -ErrorAction SilentlyContinue
 
@@ -220,6 +222,8 @@ jobs:
         uses: r-lib/actions/setup-r@v1
         with:
           r-version: ${{ matrix.config.r }}
+          # Explicit version string can be removed after
+          # https://github.com/r-lib/actions/pull/287
           rtools-version: '40v2'
 
       - name: Set up Rust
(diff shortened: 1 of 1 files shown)


Ilia-Kosenkov:
So Rtools40v2 was pushed to chocolatey and other places, and Windows images here got updated as well.
The Windows-specific job runs tests using only Rtools dependency and with no magic tricks, which (I hope) means that at least {rextendr} can work without additional dependencies.

Note that in this scenario, we use the default Rust toolchain for Windows, `stable-x86_64-pc-windows-msvc`.
It depends on [Visual Studio Build Tools](https://visualstudio.microsoft.com/downloads/#build-tools-for-visual-studio-2019), which is basically a set of native compilers, linkers and various resources.
The `stable-msvc` toolchain is then used to cross-compile to `pc-windows-gnu`, using Rtools' mingw toolset.
In short, on Windows we now use default native compilers, default native Rust toolchain, and standard Rtools distribution, and no extra configuration (apart from `PATH` I guess) is needed.

clauswilke (approved):
Looks good to me.

PR #58, opened 2021-06-06
[PoC] Automatically create pull requests to update bindings

yutannihilation:
This is a proof of concept pull request to automatically generate pull requests when there are some diffs. This runs only on master.

The result pull requests can be found on my forked repo (I merged this into its master branch): https://github.com/yutannihilation/libR-sys/pulls

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -189,6 +189,31 @@ jobs:
         env: 
           NO_TEST_TARGETS: ${{ join(matrix.config.no-test-targets, ',') }}
 
+      - name: Create a pull request if there's update to the bindings
+        if: 
+          github.ref == 'refs/heads/master' &&
+          steps.build.outcome == 'success' && 
+          steps.test.outcome == 'success' && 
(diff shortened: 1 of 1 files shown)


Ilia-Kosenkov:
I have a quick question - how would you trigger this CI once a new R version is released?
I can think of two solutions: there is a cron-like system built in GHA, allowing to run jobs on a regular basis. Never used it myself, but perhaps it can be adopted.
Or, it can be triggered by `workflow_dispatch`, i.e. manually.

Also, you need to double-check how this thing behaves once pushed to master. Because a CI triggers a push to master, I assume it should trigger all other on-push CIs, including itself (correct me if I am wrong).

So, with a separate on-demand/on-cron CI, once it is triggered, it pushes new bindings to the master, which in turn triggers regular testing CI, which then verifies if the new bindings work in all circumstances.

yutannihilation:
Thanks, good points. I'm yet to figure out the workflow using this...

> Because a CI triggers a push to master, I assume it should trigger all other on-push CIs, including itself (correct me if I am wrong).

Yes, but when there's no diff it does nothing. The problem is rather that this doesn't check if there's already the same pull request so this would end up with creating hundreds of pull requests...

PR #59, opened 2021-06-19
Fix macos-11 runner

yutannihilation:
The document says:

> macOS 11 image label was changed to macos-11 to avoid confusion with the minor version.
<https://github.com/actions/virtual-environments/blob/main/docs/macos-11-onboarding.md>

but, reading the above documentation, maybe this isn't available to us until we apply for the feature? Let's see what will happen...


yutannihilation:
Oh, this works!

PR #60, opened 2021-06-19
Setup a CI to push generated bindings to a dedicated branch

### LINKED ISSUE #57: Bindings for R 4.1
Sorry, I forgot how to update bindings. Do we need to do this? Or should the unversioned bindings (e.g. `bindings-linux-x86_64.rs`) just work?

yutannihilation:
Related to #57

This pull request adds a job to gather the uploaded generated bindings and commit them to `generated_bindings` branch. We can create a pull request from that branch at any point we want to update the bindings. Of course we can commit this to the master branch, but I feel this should be done manually with checking the diffs by our eyes, at least at the moment.

This does nothing if there's no diffs, and even if something goes wrong, this pushes to non-`master`, where no CI is set up. So, I believe this should be safe.

This pull request adds a scheduled run setting. I use this on https://github.com/ggplot2-exts/gallery for a while and it works fine.

@Ilia-Kosenkov @clauswilke 
Please let me know if there's any concern, or any topic to discuss.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -300,7 +300,6 @@ jobs:
 
         # detect changes (the code is derived from https://stackoverflow.com/a/3879077)
         git add bindings/
-
         git update-index --refresh
         if ! git diff-index --quiet HEAD -- bindings/; then
           git config --local user.name "${GITHUB_ACTOR}"
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,6 +12,8 @@ on:
   # Run this once per three days
   schedule:
     - cron: '29 17 */3 * *'
+  # This can also manually run
+  workflow_dispatch: {}
 
 jobs:
 
@@ -318,3 +320,5 @@ jobs:
         else
           echo "No changes"
         fi
+      env:
+        GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -285,8 +285,17 @@ jobs:
 
     - uses: actions/download-artifact@v2
         
-    - name: Display structure of downloaded files
+    - name: Push the generated bindings
       run: |
+        # 1) If there's already generated_bindings branch, checkout it.
+        # 2) If generated_binding branch is not created, create it from the default branch.
+        if git ls-remote --exit-code --heads upstream generated_bindings 2>&1 >/dev/null; then
+          git fetch origin --no-tags --prune --depth=1 generated_bindings
+          git checkout generated_bindings
+        else
+          git switch -c generated_bindings
+        fi
+
         # Update or add the bindings
         cp generated_binding-*/*.rs bindings/
 
@@ -305,7 +314,7 @@ jobs:
           git config --local user.name "${GITHUB_ACTOR}"
           git config --local user.email "${GITHUB_ACTOR}@users.noreply.github.com"
           git commit -m "Update bindings [skip ci]"
-          git push origin "${GITHUB_REF}":generated_bindings
+          git push origin generated_bindings
         else
           echo "No changes"
         fi
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -317,6 +317,21 @@ jobs:
           git config --local user.email "${GITHUB_ACTOR}@users.noreply.github.com"
           git commit -m "Update bindings [skip ci]"
           git push origin generated_bindings
+
+          # File an issue or a pull request
+          echo "Update the precomputed bindings"                          >  .msg
+          echo ""                                                         >> .msg
+          echo "The generated bindings have changed for these platforms:" >> .msg
+          echo ""                                                         >> .msg
+          git diff --name-only HEAD^ | sed 's/^/- /'                      >> .msg
+          echo ""                                                         >> .msg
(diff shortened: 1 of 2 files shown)


Ilia-Kosenkov:
The idea looks good. What about adding `workflow_dispatch:` trigger? This will allow us to manually force re-generation without making a PR.

Also, what if we somehow notify that there are new bindings? Is it possible to create an automatic issue in this repository? I am afraid that re-generated bindings will disappear silently in the notifications.

yutannihilation:
> What about adding `workflow_dispatch:` trigger? This will allow us to manually force re-generation without making a PR.

Thanks, sounds good. Could you explain a bit more about the latter sentence? Do you mean `workflow_dispatch` triggers a push to `master` directly? Or, is that about some new mechanism we might introduce in the future?

> Also, what if we somehow notify that there are new bindings? Is it possible to create an automatic issue in this repository?

I think it's possible. Should it create a pull request, instead of a issue, when there are new bindings? Regarding notification, it seems it's also possible to notify on Discord: https://github.com/marketplace/actions/actions-for-discord.

Ilia-Kosenkov:
> Thanks, sounds good. Could you explain a bit more about the latter sentence? Do you mean workflow_dispatch triggers a push to master directly?

No no, it is just another trigger. We trigger on PR, you add cron (every 3 days), I suggest to also have `workflow_dispatch`. That way you can trigger it directly from the Actions menu on GitHub. If there are no downsides to this, I think it is a good addition. Imagine that your scheduled run was yesterday, a new version of R is released today, and we have to wait for 2 days for bindings to re-generate or create a PR/commit to trigger the workflow. With dispatch, we can just start it when we need (in addition to on pull request and cron).

> I think it's possible. Should it create a pull request, instead of a issue, when there are new bindings? Regarding notification, it seems it's also possible to notify on Discord: https://github.com/marketplace/actions/actions-for-discord.

I was thinking about the following. If new bindings are generated, then we crate an issue saying: "The generated bindings have changed for these platforms" and some additional information (like diffs?), maybe the version of R and cargo used for generation. That way a maintainer can see there is an issue and decide what to do manually -- update bindings, publish new version to crates.io, ignore, etc.

yutannihilation:
> a PR/commit to trigger the workflow

Ah, thanks for clarifying! Sorry, I didn't explain that this runs only on master and ignores pull requests (so I was a bit confused why pull request is related here but it's totally my fault...). You can see `commit_generated_bindings` is actually skipped on this pull request.

https://github.com/extendr/libR-sys/blob/4bc89ee8208a5531de968e2e046f467172e96f71/.github/workflows/test.yml#L281-L282

Then I think I get your point. Let's add `workflow_dispatch`.

yutannihilation:
> If new bindings are generated, then we crate an issue saying: "The generated bindings have changed for these platforms" and some additional information (like diffs?), maybe the version of R and cargo used for generation.

I feel it might be a bit easier to use pull requests to include diffs, but I'll investigate both way.

yutannihilation:
Here's example issue and pull request:

* issue: https://github.com/yutannihilation/libR-sys/issues/10
* pull request: https://github.com/yutannihilation/libR-sys/pull/11

Looking at the actual diffs, I'm afraid this happens too often as there are such information like SVN revisions included. So, I'd stick with no issue or pull request for now. It's probably too complicated to handle programmatically.

Ilia-Kosenkov:
OK, let's see how it goes.

Ilia-Kosenkov (approved):

clauswilke (approved):
Fine by me. One issue we'll have to think through at some point is what to do with old bindings for older R versions that we're not covering anymore in our CI setup. Whenever a new R version is released, an old one drops out of coverage. Do we delete those bindings? Ideally we should, I think, since otherwise they can get out of sync with other changes in libR-sys.

yutannihilation:
> Do we delete those bindings?

Sorry, I forgot to answer this question on #57. I agree we should. Let's discuss this again on #57.

> It's probably too complicated to handle programmatically.

Thinking this again, maybe we can just ignore the devel bindings...? Let's revisit here when we need better mechanism to notify.

yutannihilation:
Hmm, this fails because `bindings-macos-x86_64-R4.1.rs` is generated on two macOS runners (I didn't notice this because `macOS-11` runner don't run on my folk).

```
cp: will not overwrite just-created 'bindings/bindings-macos-x86_64-R4.1.rs' with 'generated_binding-macOS-latest-R-release-rust-stable/bindings-macos-x86_64-R4.1.rs'
```

yutannihilation:
@Ilia-Kosenkov 
Do you remember why `macos-11` runner needs to check `default` target in addition to `aarch64-apple-darwin`? I briefly checked discussions around #45, but couldn't find the reason. If we can remove this, we can avoid the above conflict.

https://github.com/extendr/libR-sys/blob/8fd988b19acbc11feda80efae5c51c885a834fbf/.github/workflows/test.yml#L38-L39

I already added an workaround for this (https://github.com/extendr/libR-sys/commit/8fd988b19acbc11feda80efae5c51c885a834fbf), so this isn't very urgent, but I'd like to know the background.

Ilia-Kosenkov:
I am not sure, but I think we tested to make sure it works on the next MacOS version. I guess it can be removed. I am sorry that I did not  write any comments, at that time we were just trying to add macos11 to get arm bindings.

yutannihilation:
I see, then I'll create a pull request to update the runner. No worries.

PR #63, opened 2021-06-24
Use stable Rust to build and test aarch64-apple-darwin 

yutannihilation:
At the time of we added `macos-11` runner, the nightly toolchain was needed. But, as of [Rust 1.49](https://blog.rust-lang.org/2020/12/31/Rust-1.49.0.html), `aarch64-apple-darwin` is tier 2 target. This pull request adds the following changes regarding to this issue:

- Use stable toolchain
- ~~Remove default target~~
- ~~Remove an workaround for duplication of the generated bindings~~

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -26,6 +26,10 @@ jobs:
       fail-fast: false
       matrix:
         config:
+          # options:
+          #   - targets: Targets to build and run tests against. If not specified, 'default' will be used. 
+          #   - no-test-targets: Targets to skip tests.
+          #   - emit-bindings: If 'true', emit binding no matter which Rust version is used (by default, the bindings are emitted only with stable Rust).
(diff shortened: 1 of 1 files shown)


clauswilke (approved):

yutannihilation:
Let me figure out why the tests on aarch64-apple-darwin needed to be skipped... :thinking:

yutannihilation:
Ah, of course setup-r action installs x86_64 version because macos-11 runner is (still?) on intel machines.

> `2021-06-24T14:03:18.8608600Z   = note: ld: warning: ignoring file /Library/Frameworks/R.framework/Versions/4.1/Resources/lib/libR.dylib, building for macOS-arm64 but attempting to link with file built for macOS-x86_64`

Sorry, my attempt was a bit too early. Let's wait for this:

https://github.com/actions/virtual-environments/issues/2187

yutannihilation:
I minimized the change to just replacing the toolchain with the stable version. Let's revisit here when actions/virtual-environments#2187 gets ready...

Issue #66, opened 2021-07-18
Release libR-sys to crates.io


clauswilke:
We should make a new release to crates.io. Anything that needs to be addressed first? Maybe #31?

@yutannihilation Do you want to handle the release? I can add you as an owner on crates.io.

cc @Ilia-Kosenkov

Ilia-Kosenkov:
I experimented with a script that sets up everything needed for libR-sys compilation on a clean Windows 10 system, so I have now a better understanding of what needs to be done.

I can address #31 within the next day or so if it is ok.

clauswilke:
Super, thanks!

yutannihilation:
Sure, I have no experience of publishing a crate, but I think I can handle the release. https://github.com/extendr/libR-sys/issues/67 is nice to have, but not a blocker issue.

Thanks @Ilia-Kosenkov for addressing #31!

andy-thomason:
We should write up a "how to" in the README.

You can do a dry run at publishing the crate to find errors before committing.

We use the workspace to publish all extendr crates together on the whole. The version
numbers are usually synced.

clauswilke:
Any new headers to add before the release? 

@andy-thomason has suggested Altrep.h: https://github.com/extendr/extendr/issues/231#issuecomment-882493783

I also think `R_ext/GraphicsEngine.h` would be useful.

Ilia-Kosenkov:
https://github.com/Ilia-Kosenkov/libR-sys/tree/readme-windows
Here is my progress so far. Found an issue with the `i686` binding generation on a clean system - I am missing some dependency, will resolve it soon.

clauswilke:
@Ilia-Kosenkov You can probably simplify. I don't think we need to explain to people how to install Visual Studio.

Also, please pay attention to the structure of subsections. The following sentence is part of the Windows subsection but seems to be meant more generally.

> Once R and rust are configured, the library can be easily built:

There are a few cases like this.

Finally, I don't think we need `--test-threads=1` anymore. Do we need `-- --nocapture`?

clauswilke:
I just tried. `cargo test` is all that is needed, as far as I can tell.

```
clauswilke ~/github/libR-sys> cargo test
   Compiling libR-sys v0.2.0 (/Users/clauswilke/github/libR-sys)
    Finished test [unoptimized + debuginfo] target(s) in 3.10s
     Running target/debug/deps/libR_sys-b8d4b2aca00e72cd

running 24 tests
[...]

PR #71, opened 2021-07-22
Include R_ext/GraphicsEngine.h

yutannihilation:
Address https://github.com/extendr/libR-sys/issues/66#issuecomment-882669880


clauswilke (approved):
Thanks!

PR #76, opened 2021-08-01
Move documents about DEP_* envvars to crate level

yutannihilation:
After looking at the published doc (https://docs.rs/libR-sys/0.2.2/libR_sys/), I feel these documents should be on the crate level doc, rather than the README on crates.io.

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -30,7 +30,7 @@
 //!
 //! # Conditional compilation depending on R installation
 //!
-//! libR-sys crate provides these environmental variables that you can use in `build.rs`:
+//! The libR-sys crate provides these environmental variables that you can use in `build.rs`:
 //!
 //! - `DEP_R_R_VERSION_MAJOR`: The major part of the R version (e.g. `4` in version `4.1.0`)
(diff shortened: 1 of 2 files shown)


src/lib.rs
```diff
@@ -27,6 +27,37 @@
 //!   setup_Rmainloop();
 //! }
 //! ```
+//!
+//! # Conditional compilation depending on R installation
+//!
+//! libR-sys crate provides these environmental variables that you can use in `build.rs`:
```
clauswilke, line 33: I think this should be written as "The crate libR-sys provides ..."

clauswilke (approved):

src/lib.rs
yutannihilation: How about "The libR-sys crate provides ..."? I feel "The crate <crate name>" form seems uncommon. For example:

* "The crate serde provides": https://www.google.com/search?q="The+crate+serde+provides" (0 hit)
* "The serde crate provides": https://www.google.com/search?q="The+serde+crate+provides" (9 hits)

clauswilke: Works for me.

yutannihilation: Thanks, let's use this.

Issue #78, opened 2021-08-19
Build failure on Windows w/ i686 R-devel 


yutannihilation:
Not sure if this is reproducible, but the recent scheduled build failed. Is it time to drop support for 32-bit Windows?

```
      Compiling libR-sys v0.2.2 (D:\a\libR-sys\libR-sys)
       Running `D:\a\libR-sys\libR-sys\target\debug\build\libR-sys-1681420fd51408e7\build-script-build`
  [libR-sys 0.2.2] Problem locating local R install: Os { code: 2, kind: NotFound, message: "The system cannot find the file specified." }
  error: failed to run custom build command for `libR-sys v0.2.2 (D:\a\libR-sys\libR-sys)`
  
  Caused by:
    process didn't exit successfully: `D:\a\libR-sys\libR-sys\target\debug\build\libR-sys-1681420fd51408e7\build-script-build` (exit code: 1)
    --- stdout
    Problem locating local R install: Os { code: 2, kind: NotFound, message: "The system cannot find the file specified." }
  Error: 'Building for target: i686-pc-windows-gnu': cargo failed with code 101 (args: build -vv --features use-bindgen --target=i686-pc-windows-gnu)
```

https://github.com/extendr/libR-sys/runs/3374517306?check_suite_focus=true#step:9:2662

yutannihilation:
It seems now the default installer is 64-bit only. I guess we need to drop support for 32-bit Windows.

PR #80, opened 2021-08-22
Drop support for 32-bit Windows on R-devel

yutannihilation:
Fix https://github.com/extendr/libR-sys/issues/78, https://github.com/extendr/extendr/issues/210

Currently, the CI check fails with this error, which indicates the 32-bit version of R will be no longer installed by default as of R 4.2. This pull request drops the binding generation and delete the existing binding as well to avoid confusion. ~~This isn't the end of the world for 32-bit users as they probably can generate the bindings by themselves, though it's not very easy.~~

```
     Compiling libR-sys v0.2.2 (D:\a\libR-sys\libR-sys)
       Running `D:\a\libR-sys\libR-sys\target\debug\build\libR-sys-d9eb6d3d5a738eb6\build-script-build`
  [libR-sys 0.2.2] Problem locating local R install: Os { code: 2, kind: NotFound, message: "The system cannot find the file specified." }
  error: failed to run custom build command for `libR-sys v0.2.2 (D:\a\libR-sys\libR-sys)`
  
  Caused by:
    process didn't exit successfully: `D:\a\libR-sys\libR-sys\target\debug\build\libR-sys-d9eb6d3d5a738eb6\build-script-build` (exit code: 1)
    --- stdout
    Problem locating local R install: Os { code: 2, kind: NotFound, message: "The system cannot find the file specified." }
  Error: 'Building for target: i686-pc-windows-gnu': cargo failed with code 101 (args: build -vv --features use-bindgen --target=i686-pc-windows-gnu)
```


Ilia-Kosenkov:
According to [CRAN](https://cran.r-project.org/bin/windows/base/rdevel.html), today's release (2021-08-22) contains both 32 and 64 bit versions.
I did not try it on my machine, but I will.

Perhaps 32 bit version is not available due to some new configuration in the installation process on the CI.

Therefore for now I suggest to not throw Winx86 away, at least until we better understand what has gone wrong.

yutannihilation:
> According to [CRAN](https://cran.r-project.org/bin/windows/base/rdevel.html), today's release (2021-08-22) contains both 32 and 64 bit versions.

Oh, interesting...

A bit more context why I thought 32-bit was dropped; Two weeks ago, I found 32-bit tests fails on R-devel runners. This was because CRAN stopped distributing 32-bit binary for R-devel, but, on the other hand, `R CMD check` does check on both arch. I thought r-lib/actions need to be tweaked to make sure only the 64-bit version will be installed.

https://github.com/r-lib/actions/issues/339

But, from some days ago, R-devel runner passes, not checking 32-bit. So, I thought the 32-bit version is no longer installed.

Ilia-Kosenkov:
Oh, OK, I see.
I will later today make a local installation of devel version and check what happens under the hood.

I'd prefer to support 32-bit while it is bundled.

yutannihilation:
Thanks, I'll also check on my old Windows laptop, but it's good to double-check.

yutannihilation:
This is what I'm seeing. This is either that the installer doesn't contain 32-bit files or that it only shows the same arch as the OS. I cannot find the way to install 32-bit version.

image

Ilia-Kosenkov:
Strange, I am surprised there is no announcement and CRAN still mentions both arches. Perhaps we should indeed discard x86-devel. @clauswilke ?

yutannihilation:
I think this is the announcement, which I quoted on https://github.com/extendr/extendr/issues/210. But, yeah, it's weird that it still mentions both arches.

> It is planned that the 4.1.x series will be the last to support 32-bit Windows, with production of binary packages for that series continuing until early 2023.
(https://stat.ethz.ch/R-manual/R-devel/doc/html/NEWS.html)

Ilia-Kosenkov:
Oh yeah, you are probably right. I forgot that announcement mentioned exactly last supported version.

yutannihilation:
Honestly I too had forgotten until the CI failed :)

clauswilke:
Sorry, I'm not entirely sure what the final conclusion of the discussion is. 32 bit is unsupported from 4.2 onwards and we drop it?

clauswilke:
It'll be a bit of a mess for the package build system shipping with rextendr, but one step at a time I guess.

yutannihilation:
> 32 bit is unsupported from 4.2 onwards and we drop it?

Yes.

> It'll be a bit of a mess for the package build system shipping with rextendr, but one step at a time I guess.

The 32-bit version is no longer built or tested on R-devel runners already (c.f. r-lib/actions#339). This pull request doesn't break anything.

clauswilke:
> The 32-bit version is no longer built or tested on R-devel runners already (c.f. r-lib/actions#339). This pull request doesn't break anything.

Ah, Ok. So at some point we'll be able to simplify the makefile, but that's years in the future?

clauswilke (approved):

Ilia-Kosenkov (approved):
I support the removal then. The minor comment/request is to add a link or a couple more words (to the changelog) about why the support is dropped -- so we do not forget what was the cause and where to read about it.

yutannihilation:
> but that's years in the future?

Yeah, I guess it's 2023...

> add a link or a couple more words (to the changelog)

Sure, I will before merging this.

Issue #85, opened 2022-01-10
libR-sys fails to obtain R-version on conda-forge osx-arm64


mlondschien:
The conda-forge builds for my package (r-)changeforest fails due to
```
error: failed to run custom build command for `libR-sys v0.2.2`

Caused by:
  process didn't exit successfully: `/Users/runner/miniforge3/conda-bld/r-changeforest_1641818835367/work/changeforest-r/src/./rust/target/release/build/libR-sys-7ac94a74d09fbe03/build-script-build` (exit status: 101)
  --- stdout
  cargo:rustc-env=R_HOME=/Users/runner/miniforge3/conda-bld/r-changeforest_1641818835367/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_plac/lib/R
  cargo:r_home=/Users/runner/miniforge3/conda-bld/r-changeforest_1641818835367/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_plac/lib/R
  cargo:rustc-link-search=/Users/runner/miniforge3/conda-bld/r-changeforest_1641818835367/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_plac/lib/R/lib
  cargo:rustc-link-lib=dylib=R
  cargo:rerun-if-changed=build.rs
  cargo:rerun-if-changed=wrapper.h

  --- stderr
  thread 'main' panicked at 'Could not obtain R version: Custom { kind: Other, error: "Cannot find R major version" }', /Users/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/libR-sys-0.2.2/build.rs:393:56
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
warning: build failed, waiting for other jobs to finish...
error: build failed
make: *** [Makevars:11: rust/target/release/libchangeforest.a] Error 101
ERROR: compilation failed for package ‘changeforest’
[...]

PR #87, opened 2022-01-13
Using `R_VERSION` environement variable

Ilia-Kosenkov:
Following the discussion in #85, here I attempt to use `R_VERSION` env var to determine R version. 
I assume that `R_VERSION` will be set to something like `4.1.0`, `4.2.0-devel`, etc.

`R_VERSION` takes precedence over executing `R` to get its version to allow manual overriding through the environment variable.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -36,8 +36,10 @@ struct RVersionInfo {
 
 #[derive(Debug)]
 enum EnvVarError {
-    NotPresentOrNotUnicode,
-    Invalid,
+    EnvVarNotPresent,
+    InvalidEnvVar(&'static str),
+    RInvocationError(io::Error),
+    InvalidROutput(&'static str)
 }
 
 // frustratingly, something like the following does not exist in an
@@ -174,7 +176,7 @@ fn str_vec_to_version(input: Vec<String>) -> Result<RVersionInfo, EnvVarError> {
     }
 
     if input.len() < 3 || input.len() > 4 {
-        return Err(EnvVarError::Invalid);
+        return Err(EnvVarError::InvalidEnvVar("Invalid format"));
     }
 
     let mut result = RVersionInfo {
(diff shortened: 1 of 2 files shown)


build.rs
```diff
@@ -236,6 +291,12 @@ fn get_r_version_strings(r_paths: &InstallationPaths) -> io::Result<RVersionInfo
     })
 }
 
+fn get_r_version(r_version_env_var: &str, r_paths: &InstallationPaths) -> io::Result<RVersionInfo> {
+    get_r_version_from_env(r_version_env_var)
+    .ok_or(Error::new(ErrorKind::Other, "Cannot determine R version from environement variable"))
+    .or_else(|_| get_r_version_from_r(r_paths))
```
yutannihilation, line 297: I think the logic here should be

1. if `R_VERSION` exists and it's in the expected format, use it.
2. if `R_VERSION` exists but it's not in the expected format, return an error
3. if `R_VERSION` doesn't exist, and if the execution of `R` returns the output in the expected format, use it
4. if `R_VERSION` doesn't exist, and if the execution of `R` returns the output in the expected format (or the execution failed), return an error

I mean, the malformed `R_VERSION` should not be ignored by `or_else()`. What do you think?

Ilia-Kosenkov: Also sounds good to me. In this example the argument of `ok_or` is actually never observed. I can modify code to return something more meaningful for the case of  env var.

As an alternative, we can leave it as is but add a warning to the output stating that `R_VERSION` is malformed and therefore skipped.

yutannihilation: > As an alternative, we can leave it as is but add a warning to the output stating that `R_VERSION` is malformed and therefore skipped.

I was thinking which would be better. Are there chances that `R_VERSION` environmental variable is unintentionally set (e.g. some other application use the same name of the envvar)? If not, I think it should stop with error. The case when the user provide `R_VERSION` is probably when the auto-detection won't work correctly, so it should not fall back to the method.

Ilia-Kosenkov: I am sure it is possible that `R_VERSION` is used by someone/some toolchain -- I for instance use it locally on my machines to quickly switch environment between different R installations. But I think it can be properly set/unset before compilation if needed. I have modified the build script so now if `R_VERSION` is present but invalid, it fails, but if `R_VERSION` is missing completely, it falls back to querying R.

Ilia-Kosenkov:
I took the liberty of updating error types, I believe it should be slightly more informative now.

yutannihilation (approved):
Looks good to me. Could you also add some notes about `R_VERSION` on README?

PR #93, opened 2022-02-02
Refactor build.rs

yutannihilation:
We have added several new logics to build.rs with minimal changes. I feel now it looks fairly complex and needs to be refactored. The main changes of this pull request are the followings. The behavior of the script should stay (almost) the same.

* Currently we have two different code paths to parse the R version on envvar `LIBR_SYS_R_VERSION` and on R's output. This pull request unifies it to `parse_r_version()`.
* `probe_r_paths()` is a bit too nested. This pull request separates the logic into `get_r_home()`, `get_r_library()`, and `get_r_include()` for simplicity.
* Make the environmental variables in our convention as const and move to the top of the file so that we can discover it easier.
* Add more comments.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -336,7 +336,7 @@ fn get_r_version(
         Ok(v) => Ok(v),
         // If the envvar is not present, then use the actual R binary to get the version.
         Err(EnvVarError::EnvVarNotPresent) => get_r_version_from_r(r_paths),
-        // In the case of any other error than the absense of envvar, stop with
+        // In the case of any error other than the absense of envvar, stop with
         // that error because it means the envvar is set and something is wrong.
         e @ Err(_) => e,
     }
--- a/build.rs
+++ b/build.rs
@@ -180,7 +180,7 @@ fn parse_r_version(
     r_version: String,
     version_string: Option<String>,
 ) -> Result<RVersionInfo, EnvVarError> {
-    // Split "<major>.<minor>.<patch>-devel" to "<major>.<minor>.<patch>" and "devel"
+    // First, split "<major>.<minor>.<patch>-devel" to "<major>.<minor>.<patch>" and "devel"
     let (r_version, devel) = match *r_version.split('-').collect::<Vec<&str>>().as_slice() {
         [r_version, devel] => (r_version, Some(devel)),
         [r_version] => (r_version, None),
@@ -188,18 +188,16 @@ fn parse_r_version(
         _ => return Err(EnvVarError::InvalidEnvVar("Invalid format")),
     };
 
-    // Split "<major>.<minor>.<patch>-devel" to "<major>", "<minor>", and "<patch>"
+    // Split "<major>.<minor>.<patch>" to "<major>", "<minor>", and "<patch>"
     let r_version_split = r_version
         .split('.')
         .map(|s| {
             // Good:
             //   - "4.1.2"
-            //   - "4.2.0-devel"
             //
             // Bad:
-            //   - "4.1.foo" (contains non-digit character)
+            //   - "4.1.foo" (some part contains any non-digit characters)
             //   - "4.1." (some part is missing)
-            //   - "4.1.0-deve" (the devel part is invalid)
             if !s.is_empty() && s.chars().all(|c| c.is_digit(10)) {
                 Some(s)
             } else {
--- a/build.rs
+++ b/build.rs
@@ -128,7 +128,7 @@ fn get_r_library(r_home: &Path) -> PathBuf {
     match (cfg!(windows), pkg_target_arch.as_str()) {
         // For Windows
         (true, "x86_64") => Path::new(r_home).join("bin").join("x64"),
-        (true, "i386") => Path::new(r_home).join("bin").join("i386"),
+        (true, "x86") => Path::new(r_home).join("bin").join("i386"),
         (true, _) => panic!("Unknown architecture"),
         // For Unix-alike
         (false, _) => Path::new(r_home).join("lib"),
--- a/build.rs
+++ b/build.rs
@@ -27,7 +27,7 @@ const ENVVAR_R_HOME: &str = "R_HOME";
 // An R version (e.g., "4.1.2" or "4.2.0-devel"). When this is set, the actual R
 // binary is not executed. This might be useful in some cases of cross-compile.
 // c.f., https://github.com/extendr/libR-sys/issues/85
-const ENVVAR_R_VERSION: &str = "LIBR_SYS_R_VERSION";
+const ENVVAR_R_VERSION: &str = "LIBRSYS_R_VERSION";
 
 // A path to a dir containing pre-computed bindings (default: "bindings").
 const ENVVAR_BINDINGS_PATH: &str = "LIBRSYS_BINDINGS_PATH";
--- a/build.rs
+++ b/build.rs
@@ -1,18 +1,14 @@
 use std::{
     env,
-    ffi::OsString,
-    fs,
-    io,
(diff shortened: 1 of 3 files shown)


yutannihilation:
@Ilia-Kosenkov 
I think you felt some complexity to modify build.rs on https://github.com/extendr/libR-sys/pull/87 (as I felt so when I tried to review it...). Could you review this if my refactor is in the right direction? This isn't in a hurry, so if you have another thing in priority please ignore this for now. I can wait.

build.rs
```diff
@@ -1,22 +1,47 @@
...
+// to set manually if we compile libR-sys outside of an R session.
+//
+// c.f., https://stat.ethz.ch/R-manual/R-devel/library/base/html/EnvVar.html
+const ENVVAR_R_INCLUDE_DIR: &str = "R_INCLUDE_DIR";
+const ENVVAR_R_HOME: &str = "R_HOME";
+
+// An R version (e.g., "4.1.2" or "4.2.0-devel"). When this is set, the actual R
+// binary is not executed. This might be useful in some cases of cross-compile.
+// c.f., https://github.com/extendr/libR-sys/issues/85
+const ENVVAR_R_VERSION: &str = "LIBR_SYS_R_VERSION";
```
yutannihilation, line 30: Can we rename this to `LIBRSYS_R_VERSION` for consistency with other variables, if it's not too late...? Sorry I didn't notice this.

build.rs
```diff
@@ -75,93 +114,88 @@ fn byte_array_to_os_string(bytes: &[u8]) -> OsString {
...
+        // For Windows
+        (true, "x86_64") => Path::new(r_home).join("bin").join("x64"),
+        (true, "x86") => Path::new(r_home).join("bin").join("i386"),
+        (true, _) => panic!("Unknown architecture"),
+        // For Unix-alike
+        (false, _) => Path::new(r_home).join("lib"),
+    }
+}
+
+// Get the path to the R include either from an envvar or by executing the actual R binary.
```
Ilia-Kosenkov, line 167: '*R include folder*' or '*directory*'

build.rs
```diff
@@ -75,93 +114,88 @@ fn byte_array_to_os_string(bytes: &[u8]) -> OsString {
...
+    // Otherwise, we try to execute `R` to find `R_HOME`. Note that this is
+    // discouraged, see Section 1.6 of "Writing R Extensions"
+    // https://cran.r-project.org/doc/manuals/r-release/R-exts.html#Writing-portable-packages
+    let rout = r_command("R", r#"cat(normalizePath(R.home()))"#)?;
+    if !rout.is_empty() {
+        Ok(PathBuf::from(rout))
+    } else {
+        Err(Error::new(ErrorKind::Other, "Cannot find R home."))
+    }
+}
```
Ilia-Kosenkov, line 152: This fragment is duplicated at least twice, perhaps you can move `rout` check into `r_command()`? Or make another function that calls `r_command()` and processes output?

build.rs
```diff
@@ -75,93 +114,88 @@ fn byte_array_to_os_string(bytes: &[u8]) -> OsString {
...
-                    .to_string_lossy()
-                );
-            }
+    let rout = r_command(&r_binary, r#"cat(normalizePath(R.home('include')))"#)?;
+    if !rout.is_empty() {
+        Ok(PathBuf::from(rout))
+    } else {
+        Err(Error::new(ErrorKind::Other, "Cannot find R include."))
+    }
+}
```
Ilia-Kosenkov, line 188: Here is the second fragment.

build.rs
```diff
@@ -170,147 +204,149 @@ fn probe_r_paths() -> io::Result<InstallationPaths> {
...
     std::env::var(r_version_env_var)
         // Any error arising from reading env var is converted to this value
         .map_err(|_| EnvVarError::EnvVarNotPresent)
-        .map(|v| {
-            v.split(&['.', '-'][..])
-                .map(|s| s.to_string())
-                .collect::<Vec<String>>()
-        })
-        .and_then(str_vec_to_version)
+        .and_then(|s| parse_r_version(s, None))
```
Ilia-Kosenkov, line 274: Do we need here `None`? Is this second parameter used anywhere else?

build.rs
```diff
@@ -170,147 +204,149 @@ fn probe_r_paths() -> io::Result<InstallationPaths> {
...
-    let r_binary = if cfg!(target_os = "windows") {
-        Path::new(&r_paths.library)
-            .join("R.exe")
+    let r_binary = if cfg!(windows) {
+        Path::new(&r_paths.library).join("R.exe")
     } else {
-        Path::new(&r_paths.r_home)
-            .join("bin")
-            .join("R")
+        Path::new(&r_paths.r_home).join("bin").join("R")
```
Ilia-Kosenkov, line 281: I feel like this also repeats several times, perhaps a `fn get_r_binary(r_paths: &InstallationPaths) -> Path` would be better?

build.rs
```diff
@@ -170,147 +204,149 @@ fn probe_r_paths() -> io::Result<InstallationPaths> {
...
+) -> Result<RVersionInfo, EnvVarError> {
+    // Try looking for the envvar first.
+    match get_r_version_from_env(r_version_env_var) {
+        // If the envvar is found and it can be parsed as a valid RVersionInfo, use it.
         Ok(v) => Ok(v),
-        Err(EnvVarError::InvalidEnvVar(e)) => Err(EnvVarError::InvalidEnvVar(e)),
-        Err(_) => get_r_version_from_r(r_paths)
+        // If the envvar is not present, then use the actual R binary to get the version.
+        Err(EnvVarError::EnvVarNotPresent) => get_r_version_from_r(r_paths),
+        // In the case of any other error than the absense of envvar, stop with
```
Ilia-Kosenkov, line 332: '*any error other than*'

build.rs
```diff
@@ -170,147 +204,149 @@ fn probe_r_paths() -> io::Result<InstallationPaths> {
...
 fn set_r_version_vars(ver: &RVersionInfo) {
     println!("cargo:r_version_major={}", ver.major); // Becomes DEP_R_R_VERSION_MAJOR for clients
     println!("cargo:r_version_minor={}", ver.minor); // Becomes DEP_R_R_VERSION_MINOR for clients
     println!("cargo:r_version_patch={}", ver.patch); // Becomes DEP_R_R_VERSION_PATCH for clients
-    if ver.devel.is_empty() {
+    if ver.devel {
         println!("cargo:r_version_devel=false"); // Becomes DEP_R_R_VERSION_DEVEL for clients
     } else {
         println!("cargo:r_version_devel=true");
     }
```
Ilia-Kosenkov, line 346: Can we replace it with one `println!("cargo:r_version_devel={}", ver.devel)`?
Also, your `if` condition seems to be inverted.

build.rs
yutannihilation: Sure.

build.rs
yutannihilation: Ouch, thanks for catching...

build.rs
yutannihilation: Sounds good, thanks.

build.rs
yutannihilation: As the error messages are a bit different, I prefer to leave this.

build.rs
yutannihilation: Yeah, I too was wondering if I can remove `version_string` at all... Let's do it.

Ilia-Kosenkov (approved):
Sorry, I forgot about this PR. LGTM

yutannihilation:
Thanks for reviewing!

PR #94, opened 2022-02-05
Add description about Rtools42 for Windows R >= 4.2

yutannihilation:
This pull request addresses #396.

* Add a separate section for R >= 4.2
* Use `stable-gnu` for R >= 4.2
* Always specify the toolchain by `+` (e.g. `+stable-msvc`) for Windows because one might use different toolchain for different R project


Ilia-Kosenkov:
One important thing is once 32 bit is deprecated, we may be able to switch to a different toolchain, namely `stable-x86_64-pc-windows-gnu` and compile as-is for `x86_64-pc-windows-gnu`. Previously we observed problems when cross-compiling from 64 bit GNU to 32 bit GNU and hence are using `x86_64-msvc` to cross-compile for both arches.

This may be a significant shift in the Windows paradigm, especially when `4.2.x` is stabilized. I suggest mentioning this somewhere at the end or where you find it appropriate.

README.md
```diff
@@ -33,7 +33,7 @@ Once `R` and `rust` are configured, the library can be easily built:
   
   ```Shell
   cargo build --target x86_64-pc-windows-gnu # 64-bit
-  cargo build --target   i686-pc-windows-gnu # 32-bit
+  cargo build --target   i686-pc-windows-gnu # 32-bit, not needed for R >=4.2
```
Ilia-Kosenkov, line 36: Add space in `R >= 4.2`

README.md
```diff
@@ -44,7 +44,7 @@ Once `R` and `rust` are configured, the library can be easily built:
     ```Shell
     rustup default stable-msvc
     rustup target add x86_64-pc-windows-gnu  # 64-bit
-    rustup target add   i686-pc-windows-gnu  # 32-bit
+    rustup target add   i686-pc-windows-gnu  # 32-bit, not needed for R >=4.2
```
Ilia-Kosenkov, line 47: Add space in `R >= 4.2`

README.md
```diff
@@ -77,8 +77,24 @@ To test the build, run `cargo test`.
   ```
 - **Windows**
   
-  On Windows, both `R` and `Rtools` should be available on `PATH` for tests to succeed. Ensure that `R_HOME` points to `R` home, e.g. `C:\Program Files\R\R-4.1.0`. 
-  For `x64`, append the following to the `PATH` (using `PowerShell` syntax):
+  On Windows, first, ensure that `R_HOME` points to `R` home, e.g. `C:\Program Files\R\R-4.1.0` (on an R session, this should be set by R).
```
Ilia-Kosenkov, line 80: '*in* an R session'

README.md
```diff
@@ -77,8 +77,24 @@ To test the build, run `cargo test`.
...
+  
+  Then, also ensure `PATH` is properly configured that the following executables are available:
```
Ilia-Kosenkov, line 82: Perhaps '*Second, ensure that*' is better because you have used 'first' in the paragraph above, which implies there is a 'second' and potentially 'third' and so on.

README.md
```diff
@@ -77,8 +77,24 @@ To test the build, run `cargo test`.
...
+  
+  * the `R` binary to build against
+  * the compiler toolchain that is used for compiling the R itself, i.e., `Rtools`
+
+  Typically, these are in the following locations:
```
Ilia-Kosenkov, line 87: I propose ', *they can be found* in the following locations:'

README.md
```diff
@@ -77,8 +77,24 @@ To test the build, run `cargo test`.
...
+
+  |                  | R                         | Rtools                             |
+  | ---------------- | ------------------------- | ---------------------------------- |
+  | R <=4.1.x, 64-bit  |  `${env:R_HOME}\bin\x64`  | `${env:RTOOLS40_HOME}\mingw64\bin` |
+  | R <=4.1.x, 32-bit  |  `${env:R_HOME}\bin\i386` | `${env:RTOOLS40_HOME}\mingw32\bin` |
+  | R 4.2.x, 64-bit    |  `${env:R_HOME}\bin\x64`  | `${env:RTOOLS40_HOME}\ucrt64\bin` (this might be changed when [Rtools42] gets released) |
```
Ilia-Kosenkov, line 93: How about '*this may change once* [Rtools42] *is* released'?

README.md
```diff
@@ -77,8 +77,24 @@ To test the build, run `cargo test`.
...
+  
+  [Rtools42]: https://www.r-project.org/nosvn/winutf8/ucrt3/web/rtools.html
+
+  So, for example, if the target is 64-bit R (<=4.1.x), add the following to the `PATH` (using `PowerShell` syntax). Note that this shows how to "apeend" these for minimum side effect, but, if `PATH` already contains another version of `R` or compiler toolchain, these should be prepended to overwrite the existing ones.
```
Ilia-Kosenkov, line 97: Space in `<= 4.1.x`. I suggest modifying this paragraph and splitting it in two. First sentence should end with a colon and should be followed by the actual powershell code that shows how to modify `PATH`. After the code block, write the second sentence, but perhaps rephrase it a little bit (also there is a typo in append):

'Note that this shows how to *append paths* for minimum side effect, but if `PATH` already contains *paths to* another version of `R` *binary* or compiler toolchain, *these paths* should *instead* be prepended to *take precedence*.'.

README.md
```diff
@@ -87,7 +103,7 @@ To test the build, run `cargo test`.
   cargo test --target x86_64-pc-windows-gnu
   ```
 
-  For `x86`, 
+  For 32-bit R (<=4.1.x),
```
Ilia-Kosenkov, line 106: Here and until the end of the file -- I think when `<= 4.1.x` or similar version requirement is written with spaces around operator, it is more readable, so consider adding spaces in all such cases.

README.md
```diff
@@ -214,3 +230,10 @@ The output folder for bindings can be configured using `LIBRSYS_BINDINGS_OUTPUT_
   ```
   </details>
 
+  For 64-bit R (4.2.x), we typically need to place Rtools' UCRT toolchain before
```
Ilia-Kosenkov, line 233: Shouldn't it be  `>= 4.2.x`, as in 'all versions starting from 4.2' ?

yutannihilation:
Thanks, addressed the comments! By the way, could you consider suggesting the changes directly when the fix is simple so that I can reflect your comments quickly?

image

Ilia-Kosenkov:
Yeah, I will use the suggest edit feature more frequently, sorry.

yutannihilation:
This pull request was written when I expected Rtools40 would be used. I need to rewrite...

README.md
```diff
@@ -9,13 +9,13 @@ Low-level R library bindings
 
 ## Installation
 
-The recommended way to build this library is to use precomputed bindings, which are available for `Linux`, `macOS`, and `Windows` (`32`- and `64`-bit).
+The recommended way to build this library is to use precomputed bindings, which are available for `Linux`, `macOS`, and `Windows`.
```
Ilia-Kosenkov, line 12:
```suggestion
The recommended way to build this library is to use precompiled bindings, which are available for `Linux`, `macOS`, and `Windows`.
```

yutannihilation:
TODO:
- [ ] add `.cargo/config.toml`
- [ ] show how to tweak Rtools42 for `libgcc_eh` and `libgc_s`

yutannihilation:
Hmm, honestly I'm not sure what we should do here. I'd like to hear your thoughts.

> we can make a PR to rtools once again?

We cannot make "a PR" because Rtools42 is not on GitHub. We might be able to propose a patch on the R's Bugzilla, or make a pull request against the upstream (https://github.com/mxe/mxe), though.

> the tweaking is just copying `libgcc`

Yes, the tweaking in my mind is exactly this. But, I'm getting to feel this isn't the proper fix that is worth committing to Rtools itself. As we discussed on https://github.com/extendr/extendr/pull/376#discussion_r835887288, Jeroen left this comment on the patch of Rtools40:

```
# 'gcc' when build with --disable-shared does not produce a standalone 'libgcc_eh.a'.
# Instead, everything is put into 'libgcc.a'.
# To please compilers that hardcode -lgcc_eh, such as 'rustc', 'libgcc.a' is duplicated as 'libgcc_eh.a'.
```

So, the problem is not that there aren't `libgcc_eh` and `libgcc_s`, but that `-lgcc_eh` and `lgcc_s` are hardcoded; `rustc` requires `-lgcc_eh` and `-lgcc_s` where we can just use `-lgcc` with the compilation setting of Rtools' GCC.

I was against your idea of developing our own custom target, but I'm getting to feel you are right. What we should tweak might be Rust, not Rtools. This comes with some limitation that we should keep in mind; we'll be tied to nightly compiler and the build time will increase for compiling the core.

> Keep in mind that it is required to use a nightly compiler to build the core library.
(<https://docs.rust-embedded.org/embedonomicon/custom-target.html>)

Another option is to add Rust to MXE. There's already an open (and probably forgotten) issue: https://github.com/mxe/mxe/issues/2499. This requires more investigation, but it seems it's not impossible to add Rust. c.f.

* https://github.com/libvips/build-win64-mxe/pull/15
* https://github.com/libvips/build-win64-mxe/blob/50edec409a6bfaf4f34300966dfdb2cc131b53a3/build/plugins/llvm-mingw/patches/rust-1-fixes.patch#L83-L89

Note that this says "this patch will probably break compatibility with GCC," so we might encounter some compatibility problems (I don't understand the details yet, sorry).

So, in summary, I think we have these options:

1. Tweak Rtools42 just like Jeroen did to Rtools40
2. Tweak Rust
    a. Create a custom target
    b. Add Rust to MXE

As none of them look easy enough to finish within weeks, and probably we don't run Rust on CRAN even with Rtools42 anyway, I think the realistic solution is to ask the developers to tweak their Rtools42 on local for now.

What do you think?

Ilia-Kosenkov:
A difficult choice. Of course tweaking local installation is easy, but it makes extendr dev-only tool. Can we emit empty `libgcc` files somewhere next to the code? And remove them after compilation?

yutannihilation:
Thanks, it might be possible. I'll explore.

yutannihilation:
@Ilia-Kosenkov 
> Can we emit empty libgcc files somewhere next to the code? And remove them after compilation?

Now I'm coming to a conclusion that this is impossible; the crate can affect only the compilation of itself, but we have to inject that empty `libgcc_{eh,s}` for all the compilation including the dependency crates. Probably we can have more modest solution than modifying Rtools forever, but probably it cannot be automated. Do you think my understanding is correct?

Ilia-Kosenkov:
I guess you are correct. It seems we either can have a custom target or we can modify rtools42 as we modified rtools40. Everything else would be unreliable.

yutannihilation:
Thanks, then I'd stick with modifying Rtools for now. Let's explore if a custom target is feasible later.

yutannihilation:
Since the Windows setup is now sooo complicated, I moved the description to a dedicated section for it.

yutannihilation:
I think the rendered version should be a bit easier to read.

https://github.com/extendr/libR-sys/blob/8429921478e0bda1f08756e7d3ab2d96a56c3314/README.md#toolchain-setup-on-windows

yutannihilation:
I'll merge this in a day or two. We can improve later but I want to make minimum information available at least before the release of R 4.2.

Issue #95, opened 2022-02-10
`gamma` is not present in the bindings generated for windows


CGMossa:
After a little bit of detective work, I see that the files inside `bindings/` that contain `windows` do not have a `extern` block for the `gamma`-function, but this is present in the macos, for instance.

Any way this can be resolved?

yutannihilation:
I think it comes from this header file, which is not an R API.

https://github.com/extendr/libR-sys/blob/355cf0de0dd7f28cfa3933a70b29deb247eb2017/wrapper.h#L1

CGMossa:
Okay; I don't really know how to proceed with this.
It is unfortunate, as I wanted to use this stuff now.

clauswilke:
@CGMossa You should use a Rust-native implementation of the gamma function, e.g. [here](https://docs.rs/special/0.8.1/special/trait.Gamma.html) or [here.](https://docs.rs/statrs/0.8.0/statrs/function/gamma/index.html)

yutannihilation:
For now, lgamma is what you want.

> These functions exist for compatibility reasons. They are equivalent to lgamma etc. It is better to use lgamma since for one the name reflects better the actual computation, and moreover lgamma is standardized in ISO C99 while gamma is not. 
https://www.gnu.org/software/libc/manual/html_node/Special-Functions.html

CGMossa:
Thank you so much for this.

yutannihilation:
I made the list of actual items from R (including non-APIs). I'll create a pull request to see the diff.

https://github.com/yutannihilation/extract-R-API-items-from-header-file/blob/main/dump.txt

yutannihilation:
As the original issue seems to be resolved already (please reopen if not), I'm closing this issue. Let's file another one for #96.

PR #96, opened 2022-02-12
Use allowlist to keep the exports minimal

yutannihilation:
See https://github.com/extendr/libR-sys/issues/95#issuecomment-1037084923

Currently, libR-sys exports all symbols wildly, including those from the C standard library. I think we should keep the exports minimal to avoid confusion. This pull request tries to create a proper allowlist. The basic idea is to extract and parse the include files by ourselves first and construct an allowlist because there seems not measure to stop bindgen to parse the C standard library.


yutannihilation:
@Ilia-Kosenkov @andy-thomason 
This isn't very high priority and probably is difficult to understand, but I'd like to hear your thoughts on this. I think this is a necessary move not only to prevent such confusions as https://github.com/extendr/libR-sys/issues/95, but also to reduce the code size. Note that I've been using this branch for my CRAN package without any problem, so I believe this is ready to merge basically.

yutannihilation:
@Ilia-Kosenkov 
May I merge this?

Ilia-Kosenkov:
Looks good, but let me restart CI since the last run was ~2 weeks ago -- to be on the safe side.

Ilia-Kosenkov (approved):
LGTM, thanks & sorry for delay

yutannihilation:
No worries. Thanks for reviewing!

PR #105, opened 2022-04-23
Generated bindings for R 4.2

yutannihilation:


yutannihilation:
I needed a tweak to force to use R 4.2 (https://github.com/extendr/libR-sys/commit/e2b249da2120e5d5534c5510886a1716d76398b8) to generate this. For unknown reason, it the runner caches the response of `https://api.r-hub.io/rversions/r-release-tarball`, which is used inside the `r-lib/actions/setup-r action` to determine the version of `release`.

PR #106, opened 2022-04-29
Add an example settings.json for rust-analyzer in README

yutannihilation:


Ilia-Kosenkov:
I somehow missed this. Will you be able to upgrade it once we migrate back to `msvc` toolchain in #111.

yutannihilation:
Sure, I will.

Ilia-Kosenkov:
I encountered an issue while working with extendr where using `rtools42` required also overriding the default linker (we do this with env-var in make file). Do you know if this is also required for development?

yutannihilation:
Do you mean the setting that should go to `.cargo/config.toml`? Do you think it should go to `settings.json` instead? To be clear, to develop libR-sys itself is possible without the tweak, if I remember correctly.

https://github.com/extendr/libR-sys#tweak-the-toolchain

Ilia-Kosenkov:
Yep, that's the one. Honestly, I am not sure, I am just asking. What if you put there an env variable for linker and perhaps comment it out stating that it may be needed for some scenarios?

yutannihilation:
I too am not 100% sure, but I don't include it here because rust-analyzer is not the only one that needs the linker setting; the setting is needed on running `cargo test` etc as well, so it should go to `.cargo/config.toml` (or `.vscode/tasks.json`?). Does this make sense?

Ilia-Kosenkov:
I would say `.cargo/config.toml` is an appropriate place, I ended up tuning my environment so. Ok, let's proceed as is and then think of linker config later.

Ilia-Kosenkov (approved):

yutannihilation:
Thanks. Honestly, I'm not very good at editor things, so there might be better settings. Let's improve later.

PR #108, opened 2022-07-31
[WIP] Make winapi optional

### LINKED ISSUE #107: Can we make winapi optional?
`winapi` is needed because the output of R was not UTF-8 on Windows, and that was a problem (https://github.com/extendr/libR-sys/pull/18#issuecomment-741873561). But, it seems winapi is too heavy to include only for this purpose. I think

* The default encoding of R >= 4.2 is UTF-8 on Windows (but the default of the OS might not, so I'm not sure if it's really safe to assume that `rout` is UTF-8)
* But, at least we can set `LIBRSYS_R_VERSION` so that we don't need to communicate the output of `R`

```
$ du -sh vendor/* | grep M
5.5M    vendor/libR-sys
3.7M    vendor/libc
2.1M    vendor/syn
7.6M    vendor/winapi
53M     vendor/winapi-i686-pc-windows-gnu
55M     vendor/winapi-x86_64-pc-windows-gnu
```

yutannihilation:
Fix #107

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,12 +22,12 @@ bindgen = { version = "0.53.2", optional = true }
 regex = { version = "1", optional = true }
 
 [features]
-default = ["winapi"]
-no-r-command = []
+default = ["invoke-r-command"]
+invoke-r-command = ["winapi"]
 
 # By default, we use pre-computed bindings that ship with the library. This may fail!
 # Turn on the 'use-bindgen' feature to generate bindings on the fly for your platform.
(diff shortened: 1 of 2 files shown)


Ilia-Kosenkov:
Can we merge this?

yutannihilation:
While I opened this, I hope we don't need to merge this. Adding this option increases the cases that we need to test.

Ilia-Kosenkov:
Ok, I will open a PR with my version of these changes. Let's see how it works.

yutannihilation:
Fixed by #109

PR #109, opened 2022-08-07
Remove winapi dependency

Ilia-Kosenkov:
Following #107  and #108.

### DIFF
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ The output folder for bindings can be configured using `LIBRSYS_BINDINGS_OUTPUT_
 
   Add the following to the `PATH` (using `PowerShell` syntax). 
   ```pwsh
-  $env:PATH += ";${env:R_HOME}\bin\x64;C:\rtools42\usr\bin;C:\rtools42\x86_64-w64-mingw32.static.posix\bin;${env:MSYS_ROOT}\mingw64\bin"
+  $env:PATH = ";${env:R_HOME}\bin\x64;C:\rtools42\usr\bin;C:\rtools42\x86_64-w64-mingw32.static.posix\bin;${env:MSYS_ROOT}\mingw64\bin;${env:PATH}"
   ```
   then build & test with 
   ```pwsh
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -147,7 +147,7 @@ The output folder for bindings can be configured using `LIBRSYS_BINDINGS_OUTPUT_
 
   Add the following to the `PATH` (using `PowerShell` syntax). 
   ```pwsh
-  $env:PATH = "${env:R_HOME}\bin\x64;C:\rtools42\usr\bin;C:\rtools42\x86_64-w64-mingw32.static.posix\bin:${env:MSYS_ROOT}\mingw64\bin"
+  $env:PATH += ";${env:R_HOME}\bin\x64;C:\rtools42\usr\bin;C:\rtools42\x86_64-w64-mingw32.static.posix\bin;${env:MSYS_ROOT}\mingw64\bin"
```
yutannihilation, line 150: Thanks for catching! This was my mistake when I was changed this from prepending to appending in https://github.com/extendr/libR-sys/pull/94...

```suggestion
  $env:PATH = ";${env:R_HOME}\bin\x64;C:\rtools42\usr\bin;C:\rtools42\x86_64-w64-mingw32.static.posix\bin;${env:MSYS_ROOT}\mingw64\bin;${env:PATH}"
```

Ilia-Kosenkov: I actually used our readme to set up the environment to run tests and it failed, then I realized we had typos. Just FYI, in a typical installation scenario (several R versions, Rtools 4.0 and 4.2), it worked fine when appended.

yutannihilation: Yeah, it usually works, but sometimes doesn't. I forgot why I switched to prepend, but I guess that was to match with the way we do on the CI (appending to the `GITHUB_PATH` file means prepending to `PATH`, iiuc) and the CI was the case when things don't work well. Anyway, I'm open to suggestions. If you think it's better to revert this, feel free to file an issue!

yutannihilation (approved):
Looks good to me.

PR #113, opened 2023-01-06
Update bindings

yutannihilation:
This is a follow-up of #96; regenerate all the bindings that we actively maintain (i.e. R >= 4.1).


Ilia-Kosenkov (approved):
image
Insane change! I hope we will not break things.

yutannihilation:
Yeah, I'm 99% sure nothing will break, but I'm slightly afraid as the changes are huge...

Issue #112, opened 2022-12-29
Simplify the "emit-bindings" CI setting


yutannihilation:
I'll do after https://github.com/extendr/libR-sys/pull/111 gets merged.

https://github.com/extendr/libR-sys/blob/d91909b2086822abdc1306e9345627c6b41af78f/.github/workflows/test.yml#L200-L201

PR #114, opened 2023-01-06
Explicitly set emit-bindings: 'true'

yutannihilation:
A small improvement to close https://github.com/extendr/libR-sys/issues/112. In terms of visibility, I think it's a good idea to set `emit-bindings` explicitly, rather than detecting `stable` implicitly because the combination is getting complex...


Ilia-Kosenkov (approved):
Agreed, looks more readable now. Thanks!

PR #116, opened 2023-01-11
Add a warning when output directory for generated bindings is missing

CGMossa:
- `.cargo/config.toml` can contain environment variables for all cargo-commands.
Should we use this to set `LIBRSYS_BINDINGS_OUTPUT_PATH`? I think so.

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -1,2 +1,0 @@
-[env]
-LIBRSYS_BINDINGS_OUTPUT_PATH = { value = "bindings", relative = true }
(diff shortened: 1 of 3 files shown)


CGMossa:
The problem with the CI is that the CI is creating a `.cargo/config.toml` file by appending to an already existing one, causing there to be two `[env]` tags in the same file.
The line is:

```shell
          "@ | Out-File -FilePath .cargo/config.toml -Encoding utf8 -Append ;
```

See https://github.com/extendr/libR-sys/blob/dcc533a416f2ed35b64686b509957e62036f9798/.github/workflows/test.yml#L334

Any ideas?

Ilia-Kosenkov (approved):
Looks good, thanks!

CGMossa:
Alright, I'll merge.

PR #120, opened 2023-01-23
Use Rtools43

yutannihilation:
Close https://github.com/extendr/extendr/issues/442


Ilia-Kosenkov (approved):

yutannihilation:
Thanks for approving. As we don't see any problem related to Rtools43, I'm closing https://github.com/extendr/extendr/issues/442 by merging this. Let's revisit when we face something that needs to be addressed.

PR #122, opened 2023-01-28
Update the generated bindings

yutannihilation:
Fix https://github.com/extendr/extendr/issues/464

It is confirmed that this will fix the issue (see https://github.com/extendr/extendr/pull/465).


multimeric:
I guess you meant to link to https://github.com/extendr/extendr/pull/465?

yutannihilation:
Ah, yes. Thanks for catching...

multimeric:
If you've checked this yourself, then I'm happy to merge. The diff looks like autogenerated bindings that I don't really understand.

multimeric (approved):

yutannihilation:
Thanks! In short, different versions of the R's graphics device has different functionalities, so compilation needs to be done with the corresponding version of the header file, otherwise it fails with the error of "Graphics API version mismatch." (e.g. https://github.com/r-lib/vdiffr/issues/129)

PR #124, opened 2023-02-05
Add distribution functions

CGMossa:
This could also be useful with See extendr/extendr#474.

It will require some clever wrapping of these functions if they are to be used in Rust without a bunch of unsafe-cluttering.


ljcolling (approved):
This looks good to me. I'm already using this branch without issues.

yutannihilation (approved):

PR #127, opened 2023-02-12
Fix initials & change email to consolidate contat info between repos

Ilia-Kosenkov:
Cleaning up my name & email so it is the same across all extendr projects.


CGMossa (approved):
LGTM

Issue #132, opened 2023-02-19
Exclude "non-API" defined in R's source code


yutannihilation:
We can blocklist these symbols to prevent them from being misused. Note that, some of them are actually used in extendr-engine, so we need to exclude these from the exclusion.

https://github.com/wch/r-source/blob/8791c1706307083bc2f188a6b24fda1b5ef03cdf/src/library/tools/R/sotools.R#L376-L480

yutannihilation:
extendr-engine requires:

- `R_CStackLimit`
- `R_CleanTempDir`
- `R_RunExitFinalizers`
- `Rf_endEmbeddedR`
- `Rf_initialize_R`
- `setup_Rmainloop`

https://gist.github.com/yutannihilation/3fa914926df4ea7106cd972cd6b4f104#2-list-up-non-apis-used-in-extendr

PR #135, opened 2023-02-25
Exclude non-API calls

yutannihilation:
Close https://github.com/extendr/libR-sys/issues/132

The main change of this pull request is preventing the "non-API" calls from being included in the `allowlist`. These "non-API" are APIs that are available but prohibited to use. What does "prohibited" mean? It means `R CMD check` warns, so a package using extendr won't be accepted by CRAN. So, it's better not to expose them.

To be clear, (except that there's no guarantee that these APIs can get changed without any announcement) it's not a problem as long as the user doesn't have the ambition to submit their R package to CRAN. Actually, the extendr-engine crate some of these. So we have to keep them exported.

The list of "non-API" symbols can be found in `tools:::nonAPI`. To avoid complexity, this is dumped to a text file `nonAPI.txt` outside of `build.rs`, and periodically updated by the GitHub Actions CI; when there's any change on the list, the CI automatically creates a pull request like this: https://github.com/yutannihilation/libR-sys/pull/19.


yutannihilation:
I updated the branch using my forked repo: https://github.com/yutannihilation/libR-sys/pull/15

yutannihilation:
Hmm, strange. `free_R_HOME` and `freeRUser` were added 3 weeks ago. Is macOS's R-devel this outdated?

https://github.com/wch/r-source/blame/8c7dcc5779eb9bdebd9e26a4a0dbd26169c13a6c/src/library/tools/R/sotools.R#L462

yutannihilation:
> Is macOS's R-devel this outdated?

Oh, Intel mac seems to keep failing...

image

https://mac.r-project.org/

Anyway, I realize checking these list eagerly is unstable and not a good idea. I'll move it to a separate action just for updating nonAPI.txt.

yutannihilation:
Okay, sorry for being noisy. This should be ready now, but I don't want to merge this before publishing v0.4.0. Adding APIs should be safe, but removing might not.

I'll write the summary in description, but in short,

- The main change of this pull request is preventing the "non-API" calls from being included in the `allowlist`.
- The list of "non-API" symbols can be found in `tools:::nonAPI`. To avoid complexity, this is dumped to a text file `nonAPI.txt` and periodically updated by the CI.
    - The CI creates a pull request like this: https://github.com/yutannihilation/libR-sys/pull/19

CGMossa (approved):
If you have time to tell us why these things should be hidden, that would be great. But sure, non-api things that are exposed due to C being C, is understandable.

yutannihilation:
Thanks. 

> why these things should be hidden

To prevent tragedy like https://github.com/extendr/extendr/pull/424

PR #130, opened 2023-02-16
Add /bindings command

yutannihilation:
Close #129

This pull request allows us to generate bindings on a pull request by just commenting `/bindings`. But, this works only after this gets merged into the master branch, so it's a bit difficult to test this.

PR #136, opened 2023-02-25
Use the branch of the pull request when /bindings

yutannihilation:
When invoked by an issue comment event, the GitHub Action runs on the default branch, not on the branch of the pull request. So, we have to switch to the branch explicitly, but it's not available as a form of variables pre-defined by GitHub Actions ([doc](https://docs.github.com/en/actions/learn-github-actions/variables#default-environment-variables)). I thought I solved this problem by using `r-lib/actions/pr-fetch@v2` in https://github.com/extendr/libR-sys/pull/130, but I used this only on the step **after** building the bindings.

This pull request adds the switching step also to the `test_with_bindgen` job.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -86,6 +86,16 @@ jobs:
     steps:
       
       - uses: actions/checkout@v3
+
+      # When invoked by an issue comment event, the GitHub Actions runner runs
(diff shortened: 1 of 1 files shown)


CGMossa (approved):

yutannihilation:
Thanks!

Issue #139, opened 2023-02-26
Implement wrappers for function-like macros


multimeric:
I'm interested in accessing some reference counting functionality in R, but I noticed that e.g. `MAYBE_SHARED`, `NO_REFERENCES` have no bindings generated. This is of course because of the fact that bindgen can't (yet) generate macros for function-like macros: https://github.com/rust-lang/rust-bindgen/issues/753.

Therefore I'm thinking it might be nice to slowly add some *manual* implementations of these macros. The simple ones like `#define CHAR(x) R_CHAR(x)` are of lower priority to me since they don't represent missing functionality, but stuff like `# define MAYBE_SHARED(x) (NAMED(x) > 1)` seems useful to me, as are more complex ones like `KNOWN_SORTED`.

The only question is whether we want these to live in `libR-sys` as opposed to `extendr`, since they're not generated and would be the first manual functions in the crate. Opinions are welcome.

CGMossa:
My opinion is that these should live in `libR-sys`. I find
that providing bindings to all things R, should live in this crate.
Then others can benefit from this, or even implement alternatives to
ExtendR.
What we gain from this is more people help maintain these bindings, as
their maintenance burden is.. quite high.

yutannihilation:
A quick note. In theory, function-like macros can be caught by clang (I mean, outside of bindgen)

https://docs.rs/clang/latest/clang/struct.Entity.html#method.is_function_like_macro

but it seems there's no such symbols like `MAYBE_SHARED` here. So I have no idea how to automate this wrapper generation.

https://github.com/extendr/libR-sys/blob/183b791073288da3d37d71c6e0f461272ffb239d/build.rs#L444-L453

The version of `R_CHAR()` we can use is not a function-like macro, btw.

multimeric:
Yeah so it has hooks for when it encounters a function-like macro, but it can't automatically convert it to Rust, so we're back at the manual implementation option. You would think it possible to parse the macro like a separate preprocessor language and then convert to Rust but maybe not.

yutannihilation:
I mean, I'm slightly against the hand-crafted implementations as it sounds difficult to maintain reliably.

Just in case you might not notice, we `clang` use outside of bindgen. As it cannot distinguish symbols from the C standard library, we parse the header file and construct the allowlist that bindgen uses.

CGMossa:
I've spoken to someone and they recommend manually implementing these via `macro_rules` or even `const fn`.

I think we can do that.

[...]

PR #141, opened 2023-02-27
Gather comments from headers

CGMossa:
Adding these lines to the `clang`-call enables getting some comments.

This is not much, but it is *something*. Maybe it will be more useful later

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -484,7 +484,7 @@ fn generate_bindings(r_paths: &InstallationPaths, version_info: &RVersionInfo) {
         "Generating bindings for target: {}, os: {}, architecture: {}",
         target, target_os, target_arch
     );
-    
+
     // Point to the correct headers
     bindgen_builder = bindgen_builder.clang_args(&[
         format!("-I{}", r_paths.include.display()),
@@ -509,6 +509,9 @@ fn generate_bindings(r_paths: &InstallationPaths, version_info: &RVersionInfo) {
 
     // Finish the builder and generate the bindings.
     let bindings = bindgen_builder
+        .generate_comments(true)
+        .clang_arg("-fparse-all-comments")
+        .clang_arg("-fretain-comments-from-system-headers")
         .generate()
         // Unwrap the Result and panic on failure.
         .expect("Unable to generate bindings");
--- a/build.rs
+++ b/build.rs
@@ -591,6 +591,7 @@ fn retrieve_prebuild_bindings(version_info: &RVersionInfo) {
 #[derive(Debug)]
 struct RCallbacks;
 
+#[cfg(feature = "use-bindgen")]
 impl bindgen::callbacks::ParseCallbacks for RCallbacks {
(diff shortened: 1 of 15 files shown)


CGMossa:
This is ready for review. I just don't know who cares enough to approve / disapprove of this.

multimeric:
Ooh, I like this. Two points of feedback. One, I note that almost all of the doc comments have a leading space that at least visually looks annoying. If rustdoc is not stripping out this space then it's probably annoying enough to fix. The other thing is just that I'd like to see an extendr build run using these bindings just to run it through the tests and linting etc. Otherwise it looks good and I think it would be valuable for sure.

CGMossa:
There is an option to use `rustfmt` for the generated binding. Issue is it is completely broken for these headers.
Instead, I opted to use an instrospection thing, called `ParseCallbacks`, that allows one to edit what bindgen does in a trait-way.
I implemented a parse_comment-method;

I have investigated this to see if we had something for #139, but this parse-callback just provides you with this information as it goes through, the name of the macro, and the tokens of the macro, and you ought to do something else with it. There is no return nothing. But that's an aside.

CGMossa:
For the other point, see https://github.com/extendr/extendr/pull/497, once it finishes whatever it needs to do.

CGMossa:
Okay, so the answer is, that this `[redacted].clang_arg("-fretain-comments-from-system-headers")` is the problem.
I don't know why, and I cannot check this with a Mac machine.
I guess it could be a todo for someone with a Mac.

CGMossa:
Alright; I'm re-running everything, but it already worked out, and I've addressed comments from you @multimeric.
PS: I wrote this message on the wrong PR.
But this one is ready for round two.

multimeric (approved):
Looks great!

build.rs
```diff
@@ -508,18 +509,24 @@ fn generate_bindings(r_paths: &InstallationPaths, version_info: &RVersionInfo) {
 
     // Finish the builder and generate the bindings.
     let bindings = bindgen_builder
+        .generate_comments(true)
+        .parse_callbacks(Box::new(RCallbacks))
+        .clang_arg("-fparse-all-comments")
+        // .clang_arg("-fretain-comments-from-system-headers")
```
yutannihilation, line 515: Do we need this line?

```suggestion
```

yutannihilation (approved):
Looks good to me!

CGMossa:
No, but https://github.com/extendr/libR-sys/pull/141#issuecomment-1448804819

CGMossa:
It didn't do anything on the windows part.
But it clearly affects the other platforms. So there could be something
there...

I'll remove it (or either of you can remove the line) and then let's merge
this.
It is not clean to have commented lines or code anyways.

multimeric:
Urgh I just remembered the changelog. Please add a PR updating it too.

yutannihilation:
Thanks! Putting it behind `#[cfg(not(target_os = "macos"))]` can be an option, but I'm fine with either.

Issue #144, opened 2023-03-05
`VECTOR_PTR` doesn't return on some platforms


CGMossa:
On _some_ platforms, `VECTOR_PTR` returns `!` which means it never comes back again.

This is actually a cause for concern.
Can anyone using these platforms do any digging?
Maybe @yutannihilation you might know something? 

Here you see the signatures from the different binding files, so you can see if your platform is one of the never
ending ones.

<details>
```
25 results - 25 files

bindings\bindings-linux-aarch64-R4.1.rs:
  1179  extern "C" {
  1180:     pub fn VECTOR_PTR(x: SEXP) -> !;
  1181  }

bindings\bindings-linux-aarch64-R4.2.rs:
  1223  extern "C" {
  1224:     pub fn VECTOR_PTR(x: SEXP) -> !;
  1225  }

bindings\bindings-linux-aarch64-R4.3-devel.rs:
  1228  extern "C" {
  1229:     pub fn VECTOR_PTR(x: SEXP) -> *mut SEXP;
  1230  }

bindings\bindings-linux-x86_64-R3.6.rs:
  3402  extern "C" {
  3403:     pub fn VECTOR_PTR(x: SEXP) -> *mut SEXP;
  3404  }

bindings\bindings-linux-x86_64-R4.0.rs:
  3431  extern "C" {
  3432:     pub fn VECTOR_PTR(x: SEXP) -> *mut SEXP;
  3433  }

bindings\bindings-linux-x86_64-R4.1.rs:
  1179  extern "C" {
  1180:     pub fn VECTOR_PTR(x: SEXP) -> !;
  1181  }

bindings\bindings-linux-x86_64-R4.2.rs:
  1223  extern "C" {
  1224:     pub fn VECTOR_PTR(x: SEXP) -> !;
  1225  }

bindings\bindings-linux-x86_64-R4.3-devel.rs:
  1228  extern "C" {
  1229:     pub fn VECTOR_PTR(x: SEXP) -> *mut SEXP;
  1230  }

bindings\bindings-macos-aarch64-R4.0.rs:
  3159  extern "C" {
  3160:     pub fn VECTOR_PTR(x: SEXP) -> *mut SEXP;
  3161  }

bindings\bindings-macos-aarch64-R4.1.rs:
  3046  extern "C" {
  3047:     pub fn VECTOR_PTR(x: SEXP) -> *mut SEXP;
  3048  }

bindings\bindings-macos-aarch64-R4.2.rs:
[...]

PR #145, opened 2023-03-05
Upgrade `bindgen`

CGMossa:
Only decided to do this, to hope that it solves #144.


CGMossa:
Alright. It is not possible to specify Clang version (through the `clang`-crate's features) and have it work at all.
Nor is it easy to embed the version of clang used in the generation of bindings. There is a function for it in `clang::get_version()` that returns a string, but putting that in the binding files turns out to be a pain. I'm closing this for now.

PR #148, opened 2023-03-10
Define NORET manually

yutannihilation:
Note: I chose `__attribute__((__noreturn__))` instead of `__attribute__((noreturn))` as this is suggested on the following SO.

https://stackoverflow.com/a/28556713


CGMossa (approved):
I haven't checked if this ensures all bindings have the never type return on noret, but I'm sure this is a step in the right direction anyways.

yutannihilation:
Thanks.

PR #150, opened 2023-04-02
Address the change of Rcomplex in R 4.3? 4.4?

yutannihilation:
`build.rs` didn't expect `Union` will be used in the API, so this pull request just adds it.

Note that, while the fix is very simple, I think extendr will no longer compile because it assumes `Rcomplex` is a struct.


Ilia-Kosenkov (approved):

kevinushey:
Can I ask, are there plans to push a new release of `libR-sys` to crates.io with this change?

kevinushey:
As I understand it, the current release of `libR-sys` on crates.io cannot be used with R 4.3? We could install an older version of R, but that gets a bit hairy (especially on GitHub Actions where the version of R installed by default is R 4.3)

CGMossa:
You can link to the github version directly from your rust project.

yutannihilation:
@kevinushey 
Sorry for my laziness. Yes, I was thinking I should publish a new version.

Just to be clear, libR-sys ended up with overwriting the `Rcomplex` definition to the old one before R 4.2, so you should not have any problem if you are using the precomputed bindings: https://github.com/extendr/libR-sys/pull/156/files#diff-deb55fa2ae44b6a308358532a2ee3b866bd5866626cad96c31108d2f13d2f9d3

But, if you generate the bindings by yourself, you'll definitely need this change.

yutannihilation:
@kevinushey
Do you see any actual problem related to Rcomplex because of not releasing the new version? I expect there's nothing, but If there's any, it means I might not understand the problem correctly. So, I'm curious.

kevinushey:
Sorry, I missed your message! We are all set on our side now, though -- thank you :-)

yutannihilation:
Thanks for sharing!

PR #153, opened 2023-04-09
Rework CI

### LINKED ISSUE #152: Replace `MSYS2` with native `LLVM` + `RTools`
While investigating #151 , I managed to generate bindings using native `LLVM` installation without relying on `MSYS2`. I used `RTools` to provide C/C++ standard headers (`LLVM` does not bundle it apparently), and it worked like a charm. It even worked for old version of R, I was able to generate 32-bit bindings using for R 3.6.3. 
The requirements are:
- `choco` installation of `LLVM` (which sets up path correctly)
- `mingw gcc` on path (`Rtools/blah-blah/bin`)
  - if we want to use `ucrt` version, then we need a `.cargo/config` and linker for `x86_64` set to `x86_64-w64-mingw32.static.posix-gcc.exe`
- `LIBRSYS_LIBCLANG_INCLUDE_PATH` to point to standard `C/C++` headers (also, `Rtools/blah-blah/include`)

Here `blah-blah` is `x86_64-w64-mingw32.static.posix` when a newer version of `Rtools` & 64-bit only R are used, and `mingw64` for older versions, and `mingw32` if we target 32-bit.

Ilia-Kosenkov:
This large PR attempts to accomplish two things:
- Migrate from `MSYS2` to native `LLVM` installation when generating bindings for Windows
- Adopt 'one target - one pipeline' strategy to reduce the complexity of the CI script

Closes #152.


Ilia-Kosenkov:
The `/ bindings` command does not seem to work as expected since it is triggered using the pipeline from `master` branch. I can in principle download pipeline artifacts and compute diff with the existing bindings to see if there are any discrepancies.

yutannihilation:
> The `/ bindings` command does not seem to work as expected

What do you mean? It just ran and failed because you don't merge the current master into this yet.

https://github.com/extendr/libR-sys/actions/runs/4679327017

yutannihilation:
Ah, okay, I think I understand what you meant. Yes, it runs against the current branch, not the version of `master` merged with this branch. I configured so because otherwise all the commits from master will be pushed, messing up the diffs.

.github/workflows/test.yml
```diff
@@ -109,55 +116,55 @@ jobs:
...
+      - name: Configure Windows
+        if: runner.os == 'Windows'
         run: |
-          if ($env:RUST_TARGETS -eq '') {
-            $env:RUST_TARGETS = "default"
-          }
-          foreach ($target in ($env:RUST_TARGETS).Split(",")) {
-            if ($target -ne "default") {
-              rustup target add $target
+          choco install llvm --version 15.0.7 -y
```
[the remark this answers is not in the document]
Ilia-Kosenkov: Windows runner currently [has `LLVM 15.0.7`](https://github.com/actions/runner-images/blob/main/images/win/Windows2022-Readme.md) so I suggest we force-install `16.0.1`

Ilia-Kosenkov:
I found a misconfiguration in my CI and I also started having issues with R-oldrel binding generation for 32-bit. I need more time for investigation.

.github/workflows/test.yml
```diff
@@ -109,55 +116,59 @@ jobs:
...
+      - name: Configure Windows
+        if: runner.os == 'Windows'
         run: |
-          if ($env:RUST_TARGETS -eq '') {
-            $env:RUST_TARGETS = "default"
-          }
-          foreach ($target in ($env:RUST_TARGETS).Split(",")) {
-            if ($target -ne "default") {
-              rustup target add $target
+          choco install llvm -y --version 16.0.1 --force
```
yutannihilation, line 131: Sorry to comment here again. Why do we need a specific version of LLVM? LLVM is a bit heavy to install (lighter than installing R, though), so I think it's nice if we can skip this.

.github/workflows/test.yml
```diff
@@ -55,25 +55,32 @@ jobs:
...
-          - {os: windows-latest, r: 'release', rust-version: 'stable-gnu',   targets: ['x86_64-pc-windows-gnu'], rtools-version: '42'}
-          - {os: windows-latest, r: 'oldrel',  rust-version: 'stable-msvc',  targets: ['x86_64-pc-windows-gnu', 'i686-pc-windows-gnu'], emit-bindings: 'true'}
+          - {os: windows-latest, r: 'release', rust-version: 'stable-msvc',  target: 'x86_64-pc-windows-gnu', rtools-version: '42', emit-bindings: 'true'}
+          - {os: windows-latest, r: 'release', rust-version: 'nightly-msvc', target: 'x86_64-pc-windows-gnu', rtools-version: '42'}
+          - {os: windows-latest, r: 'devel',   rust-version: 'stable-msvc',  target: 'x86_64-pc-windows-gnu', rtools-version: '42', emit-bindings: 'true'}
+          - {os: windows-latest, r: 'release', rust-version: 'stable-gnu',   target: 'x86_64-pc-windows-gnu', rtools-version: '42'}
+          - {os: windows-latest, r: 'oldrel',  rust-version: 'stable-msvc',  target: 'x86_64-pc-windows-gnu', rtools-version: '40', emit-bindings: 'true'}
+          - {os: windows-latest, r: 'oldrel',  rust-version: 'stable-msvc',  target: 'i686-pc-windows-gnu', rtools-version: '40', emit-bindings: 'true'}
```
yutannihilation, line 63:
```suggestion
          - {os: windows-latest, r: 'release', rust-version: 'stable-msvc',  target: 'x86_64-pc-windows-gnu', rtools-version: '42', emit-bindings: 'true'}
          - {os: windows-latest, r: 'release', rust-version: 'nightly-msvc', target: 'x86_64-pc-windows-gnu', rtools-version: '42'}
          - {os: windows-latest, r: 'devel',   rust-version: 'stable-msvc',  target: 'x86_64-pc-windows-gnu', emit-bindings: 'true'}
          - {os: windows-latest, r: 'release', rust-version: 'stable-gnu',   target: 'x86_64-pc-windows-gnu', rtools-version: '42'}
          - {os: windows-latest, r: 'oldrel',  rust-version: 'stable-msvc',  target: 'x86_64-pc-windows-gnu', emit-bindings: 'true'}
          - {os: windows-latest, r: 'oldrel',  rust-version: 'stable-msvc',  target: 'i686-pc-windows-gnu', emit-bindings: 'true'}
```

We only need to set `rtools-version` for R 4.2 because it's the only version where two Rtools (40 and 42) can be used. Setting `43` explicitly was my fault, sorry.

cf., https://github.com/r-lib/actions/commit/740a0e7ce930b6357f16388c0b2acee8337adb01

If you still think explicitness is good, I'm fine then, but please fix the `rtools-version` for R-devel to `43`.

.github/workflows/test.yml
Ilia-Kosenkov: Ok, but we will be using 'older' version of LLVM compared to MSYS2

.github/workflows/test.yml
Ilia-Kosenkov: The logic that I employ actually requires `rtools-version`. One way to obtain it is to use output from a previous step. Let me check if this can be done here.

Ilia-Kosenkov: The logic that I employ actually requires `rtools-version`. One way to obtain it is to use output from a previous step. However, I just checked that R installation action does not output rtools version https://github.com/r-lib/actions/blob/4723ad22e584161241adeffe987de81ddec698b6/setup-r/action.yml#L54-L56. I do not want to try & guess which version of Rtools should be used, so let us keep explicit version numbers in the CI setup.

yutannihilation: I see

yutannihilation (approved):
Looks good!

Ilia-Kosenkov:
Since it is a significant change, @CGMossa, please also take a look at this before we merge.

CGMossa:
Looks good. I've been using this setup locally ever since you mentioned it.
I'm no CI expert so my opinion is a little rudimentary on this one.
This still contains the issue that bindings aren't copies a over, but @yutannihilation is dealing with that in a separate PR.

LGTM

CGMossa:
If you can answer the two questions I put up. That would be nice

Ilia-Kosenkov:
I am sorry I think I lost my focus. Which two questions are you referring to?

PR #154, opened 2023-04-10
Maintenance PR + Hotfix

CGMossa:
UPDATE: This PR got a little out of hand, so here's a description on what it does.

- `get_non_api` is only used when `use-bindgen` is enabled.
- Rename `RCallbacks` to reflect what they actually do right now, which is trimming the comments processed by clang, i.e. `TrimCommentsCallbacks`. 
- BUG FIX: On Windows paths to dylib/dlls don't get processed properly. Added a `.canonicalize` and it seems to fix it.
- Updated bindgen to latest version `0.64`. 
- Added a versioning header to all bindings; e.g.
```
/* bindgen clang version: Ubuntu clang version 14.0.0-1ubuntu1 */
/* clang-rs version: Ubuntu clang version 14.0.0-1ubuntu1 */
/* r version: 4.1.3 */
/* automatically generated by rust-bindgen 0.64.0 */
[...]
````
- BUG FIX: Added a `is_anonymous` check on the entities processed. This is necessary as some unamed items
gets a name in llvm16, and that isn't handled in any meaningful way by `clang-rs` unfortunately. This requires `clang_3_7` atleast, which is also added.
- Updated bindings

I spent a lot of time inside the `build.rs` file and found a few things.
See the commit descriptions as well.

- [x] Ensure generated bindings work with `extendr/extendr`. (i.e. no missing stuff please!)

### CODE REVIEW

**Comment by CGMossa** [comment]
I've ended up adding a hotfix for the issue that @Ilia-Kosenkov have been hunting down.
We've reported this to https://github.com/rust-lang/rust-bindgen/issues/2488, and then I found out that it could potentially be the addition of unnamed structs in the named list of items in `allowlist_pattern`.

Needless to say, this should be removed in the future of course, once we get things patched upstream

**Comment by CGMossa** [comment]
\bindings

**Comment by CGMossa** [comment]
Temporarily fixes https://github.com/extendr/libR-sys/issues/151

**Comment by yutannihilation** [comment]
Thanks for the great investigation! But, I see suspicious diffs on the bindings. I have no idea why the CI tests won't fail. Any ideas?

**Comment by CGMossa** [comment]
No I don't. But I don't like this either.
I will export the `allowlist_pattern` lists and investigate if the
allowlist suddenly changed and in turn no longer contains the "good" stuff.

Unfortunately, this is somewhat cumbersome to do. But I'll get to it.

Do you know if the `/bindings` command invokes CI? Why not?

**Comment by yutannihilation** [comment]
> Do you know if the `/bindings` command invokes CI? Why not?

It doesn't invoke the CI because the commit message contains "[skip ci]".
The intent of this is that, as the bindings are the result of the CI runs, it should have failed if something is wrong.

And, sorry, I thought the CI checks with extendr so it would fail, but I was wrong here.

**Comment by CGMossa** [comment]
While this passes CI, running `extendr` on this PR doesn't pass. Example is `R_BaseEnv` is needed, but it got deleted (for some configurtion).

Thus I need to investigate this.. Sorry for the inconvenience.

**Comment by CGMossa** [comment]
Unfortunately, I do not have a path forward. Running this branch locally produces the bindings without the missing elements (atleast the one I sit on).
I can also not find where the logs for generating the bindings is at, so I'm unable to inspect what might have went on on the runner.

I definitely need assistance on this :(

**Comment by CGMossa** [comment]
No matter what I do, I'm unable to replicate the generated file on my machine with the one provided by the CI.

My version and the CI version agree on these

```
/* bindgen clang version: clang version 16.0.0 */
/* clang-rs version: clang version 16.0.0 */
/* r version: 4.2.3 */
/* automatically generated by rust-bindgen 0.64.0 */
```

and still there are some items that appear in my generated bindings locally, that are not present in the CI.

**Comment by yutannihilation** [comment]
Ah, so maybe adding `"` made the items on the head and tail of the allowlist unmatched?

https://github.com/extendr/libR-sys/pull/154/commits/b59b60159de70cb8c464b5ab7c9b1a3b79f21f7d

**Comment by CGMossa** [comment]
I lost a lot of my day yesterday to this little... issue.
I've updated the PR description and I kindly ask you to review and approve so this can be dealt with for now.

I think what we have right now is as much as an official fix will look like. Not just a hotfix.

File: build.rs [hunk 1]
```diff
@@ -445,7 +448,14 @@ fn generate_bindings(r_paths: &InstallationPaths, version_info: &RVersionInfo) {
         match e.get_kind() {
             EnumDecl | FunctionDecl | StructDecl | TypedefDecl | VarDecl | UnionDecl => {
                 if let Some(n) = e.get_name() {
-                    allowlist.insert(n);
+                    // due to a bug in LLVM 16.0.0 on Windows, unnamed
+                    // items are included here
+                    // See
+                    // https://github.com/rust-lang/rust-bindgen/issues/2488
+                    // https://github.com/extendr/libR-sys/issues/151
+                    if !e.is_anonymous() {
```
Ilia-Kosenkov, line 456: Should we move this check out of match expression?

build.rs
```diff
@@ -445,7 +448,14 @@ fn generate_bindings(r_paths: &InstallationPaths, version_info: &RVersionInfo) {
         match e.get_kind() {
             EnumDecl | FunctionDecl | StructDecl | TypedefDecl | VarDecl | UnionDecl => {
                 if let Some(n) = e.get_name() {
-                    allowlist.insert(n);
+                    // due to a bug in LLVM 16.0.0 on Windows, unnamed
+                    // items are included here
+                    // See
+                    // https://github.com/rust-lang/rust-bindgen/issues/2488
+                    // https://github.com/extendr/libR-sys/issues/151
```
Ilia-Kosenkov, line 455: There were some clarifications & explanation sin the bindgen issue, can you update these comments accordingly?

Ilia-Kosenkov (approved):
Amazing job, really well done! Thanks for also catching some annoying bugs, I'll definitely check it locally, especially whether the search  path issue is resolved on Windows.

Feel free to resolve or reject the comments I left, they are mostly stylistic

build.rs
CGMossa: I moved it. But I don't know if the new location is better.

build.rs
CGMossa: I don't know exactly what to retain. Honestly, I'm just going to reference your/our issue and just call it a day there. The use of `Entity::is_anonymous` is mentioned there.

bindings/bindings-linux-aarch64-R4.4-devel.rs
```diff
@@ -639,7 +639,6 @@ pub const Rboolean_FALSE: Rboolean = 0;
 #[doc = ", MAYBE"]
 pub const Rboolean_TRUE: Rboolean = 1;
 pub type Rboolean = ::std::os::raw::c_uint;
-#[doc = "This definition uses an anonymous structure, which is defined in C11 (but\nnot C99).  It is, however, supported at least by GCC, clang and icc.  The\nprivate_data_c member should never be used in code, but tells the compiler\nabout type punning when accessing the .r and .i elements, so is safer to use\nwhen interfacing with Fortran COMPLEX*16 or directly C99 _Complex double\n(PR#18430).\n\nThis form of static initialization works with both definitions:\nRcomplex z = { .r = 1, .i = 2 };\n\nAnonymous structures and C99 _Complex were not incorporated into C++\nstandard.  While they are usually supported as compiler extensions, warnings\nare typically issued (-pedantic) by a C++ compiler."]
```
Ilia-Kosenkov: Why is this now gone from some of the bindings? Is it because the version has changed?

CGMossa: For reference:
https://github.com/r-devel/r-svn/blob/master/src/include/R_ext/Complex.h

See the history. 

This is unfortunately not an exact science, and these comments are not put there to be found my automatic tools like `bindgen`, so if they disappear, I cannot but be a little 🤷‍♂️ about it.

CGMossa:
@Ilia-Kosenkov can you confirm that I may merge this?

yutannihilation (approved):
Great work, thanks!

PR #155, opened 2023-04-12
Temporary fix for Rcomplex

CGMossa:
I propose we merge this temporary fix for https://github.com/extendr/extendr/issues/524, just so that we move forward with a functioning CI.


Ilia-Kosenkov (approved):
Ok, makes sense. Unless they spontaneously decide to remove this option in one of the next versions :-)

PR #156, opened 2023-04-12
Redefine R-devel's `Rcomplex`

CGMossa:
Instead of fighting everything, we can just provide functions that help us manipulate `Rcomplex` through C99 / C11.

Depends on:
- [ ] #157 to pass CI.


CGMossa:
@yutannihilation do you know how I could solve this? I need the generated bindings before this can pass ci.

yutannihilation:
I can't see what's happening here... Do you mean you can generate bindings on your local?

I guess you have to add `create_rcomplex` (and other functions as well?) to `allowlist`.

CGMossa:
If I run `cargo build --features use-bindgen` first, then a bindings are generated, and then calls like `cargo test` work without a problem.

yutannihilation:
Then, I have no idea. I might have time to look into at night, sorry.

CGMossa:
Yeah no problem. I wouldn't ask for more. Thanks for attempting to solve it quickly.

yutannihilation:
Sorry, now I understand what you meant. I think the current CI setup is wrong. This says "Run tests on precomputed bindings shipped with libR-sys," but I guess the intent of this is to test against the new bindings generated on the previous step. This is probably my mistake when I tweaked the settings to upload the bindings as artifacts.

https://github.com/extendr/libR-sys/blob/df48d3974e89aaccbaf8c53e22b6f2f9b0a772cf/.github/workflows/test.yml#L247-L260

CGMossa:
Repurposing this pull-request. @Ilia-Kosenkov suggested a replacement in the wrapper. On my end, it looks like it works. Thus I'll push that. Hopefully CI will be sorted and this can be reviewed+merged soon.

CGMossa:
CI passes because now we aren't really requiring much from it. The changes @yutannihilation mentioned still need to happen.

This works on ExtendR locally, but it needs to be merged for us to know.

I could add tests to this. They should atleast include an array test for this.

wrapper.h
```diff
@@ -43,3 +40,11 @@ typedef ptrdiff_t R_xlen_t_rust;
 #include <R_ext/GraphicsEngine.h>
 #include <R_ext/Applic.h>
 #include <R_ext/Random.h>
+
+
+/// <div rustbindgen replaces="Rcomplex"></div>
```
Ilia-Kosenkov, line 45: Can you write a regular comment above this explaining why we re-define `Rcomplex`, perhaps with some links? Similar to what is written before `R_xlen_t`.

CGMossa:
Come on! Suggest the comment. It was your idea.

Ilia-Kosenkov (approved):
LGTM. The conflicts are due to the new version of R-devel, so just accept whatever changes your PR introduces over existing files.

yutannihilation:
Oh, oldrel is already switched to R 4.2, which means it will never pass until we update the CI setting. You can ignore the CI failure for now.

yutannihilation:
I'd like to regenerate the bindings on master to slide the version, so let me merge this before that.

PR #159, opened 2023-04-15
Refactor version logging in generated bindings

CGMossa:
It is very helpful to have the information of which clang version was used,
and which R version was used in the generated bindings.

* Unfortunately I introduced this in a very verbose way, when there are
facilities to include text lines in the generated bindings.

* I've also cleaned up a little bit the `build.rs`. 

I feel like this script is touched by us maintainers all the time, and
will be a constant nuisance, so I do vote we make it as easy to work with,
and as clear as possible. That's the spirit behind this PR.


build.rs
```diff
@@ -547,6 +551,12 @@ fn generate_bindings(r_paths: &InstallationPaths, version_info: &RVersionInfo) {
 
     // Finish the builder and generate the bindings.
     let bindings = bindgen_builder
+        .raw_line(format!(
+            "/* bindgen clang version: {} */",
+            bindgen::clang_version().full
+        ))
+        .raw_line(format!("/* clang-rs version: {} */", clang::get_version()))
+        .raw_line(format!("/* r version: {} */", version_info.full))
```
Ilia-Kosenkov, line 559: Should we also write `libR-sys` version?

CGMossa: Honestly, that's a very good idea.. I'd also like the `bindgen` version.

CGMossa: Added the `libR-sys` version, but the `bindgen`-dependency version is not possible for me atleast. See https://github.com/rust-lang/cargo/issues/7846 if you're curious. I thought the `DEP_BINDGEN_PKG_VERSION` would work.

Ilia-Kosenkov (approved):
I think this one is also safe to merge.

PR #160, opened 2023-04-15
Use rust constants and rust types in bindgen

CGMossa:
In continuing the effort to clean-up and smooth out libR-sys, I have found that clippy complains about these constants, and we can replace c-types with rust types by this setting.

This will need to be checked more thoroughly with extendr though.

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -507,6 +507,7 @@ fn generate_bindings(r_paths: &InstallationPaths, version_info: &RVersionInfo) {
         .allowlist_function(&allowlist_pattern)
         .allowlist_var(&allowlist_pattern)
         .allowlist_type(&allowlist_pattern)
+        .translate_enum_integer_types(true)
         // The input header we would like to generate
         // bindings for.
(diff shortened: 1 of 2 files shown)


CGMossa:
This is also blocked by #157 
- [ ] #157

yutannihilation:
Thanks for catching. I think we can just block them. Are there any good reason to re-export these constants that are already available via std?

CGMossa:
Maybe extendr uses the rust name?
This yields documentation through the rust version atleast.

yutannihilation:
> Maybe extendr uses the rust name?

I quickly searched `M_` in the extendr's repository, but nothing is found. So, I think it's fine to remove.

CGMossa:
That's even better. I'll make adjustments.

CGMossa:
I've removed them.

CGMossa:
We could consider `.blocklist_item("^M_.*")` just to clean up.

yutannihilation (approved):
Looks good!

PR #161, opened 2023-04-16
Find macros in R-headers using clang

CGMossa:
Previously we had a scary regex to find macros.
I found out that if you ask `clang-rs`'s `Parser` with `.detailed_preprocessing_record(true)` you get the macros as well.

This simplifies the `build.rs` substantially.


yutannihilation (approved):
Wow, I didn't know `is_in_system_header()`. Looks great!

PR #165, opened 2023-04-27
Remove `VECTOR_PTR` from bindings

CGMossa:
This should have happened with https://github.com/extendr/libR-sys/issues/144. 

Best source I can found on the deprecated status of `VECTOR_PTR` (except the r-source in #144), is
> (There is STRING_PTR() which is used in a handful of places in the R source; VECTOR_PTR() is a deprecated interface that now throws an error.)

in https://github.com/hadley/r-internals/blob/master/vectors.md


yutannihilation (approved):
Looks good.

PR #169, opened 2023-05-27
Bump version to v0.5.0 and add changelog

yutannihilation:


yutannihilation:
/bindings

(I need to run this to update the versions embedded in the generated bindings)

Ilia-Kosenkov (approved):
Looks good to me!

andy-thomason:
If you need a fresh crates.io token, I will send one.

I'm going to be in Belgrade all week, but I will be working.

yutannihilation:
Thanks, but multiple members are already the owner of this crate, so no worries.

https://crates.io/crates/libR-sys

image

yutannihilation:
@CGMossa 
No hurry, but could you take a look on the changelog? As most of the changes were your contribution, I'd like to hear your comments.

CHANGELOG.md
```diff
@@ -1,8 +1,14 @@
 # Changelog
 
-## libR-sys (unreleased)
+## libR-sys 0.5.0
 
+- Update the default precomuted bindings to R 4.3.0.
```
CGMossa, line 5:
```suggestion
- Update the default precomputed bindings to R 4.3.0.
```

CHANGELOG.md
```diff
@@ -1,8 +1,14 @@
...
+- Remove deprecated API `VECTOR_PTR` from bindings [[#165]](https://github.com/extendr/libR-sys/pull/165)
+- Use Rust constants (i.e., `::std::os::raw::c_uint` is now replaced with `u32`) [[#160]](https://github.com/extendr/libR-sys/pull/160)
```
CGMossa, line 7: Use rust types instead of c-types. I snuck that in there because R don't use exotic types across all the supported platforms. I think.. And thus I snuck it in there with the constants removal.

CGMossa (approved):
Thanks for pinging me on this!
Only thing that needs to be changed is the first "use rust constants" comment.
I added [`translate_enum_integer_types`](https://docs.rs/bindgen/latest/bindgen/struct.Builder.html#method.translate_enum_integer_types), which **I think** means that the enums/variants/discriminants that are used to indicate R-types in C, instead of being `c_int`, or `c_uint` are not `i32` or `u32` resp.

CHANGELOG.md
yutannihilation: Thanks for catching!

CHANGELOG.md
CGMossa: That's much better than whatever I wrote. Perfect!

yutannihilation: Thanks!

PR #170, opened 2023-06-30
Remove 32-bit; remove support for R < 4.2

yutannihilation:
Part of https://github.com/extendr/extendr/issues/210. It might be a bit too early, but it already passed one year and several months.

Dropping 32-bit means R < 4.2 on Windows no longer works, so this pull request drops bindings for R < 4.2 as well for consistency.


Ilia-Kosenkov (approved):

yutannihilation:
Thanks.

PR #173, opened 2023-07-10
Allow non-existent path of r_library

yutannihilation:
Address https://github.com/extendr/extendr/issues/580

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -194,13 +194,23 @@ fn get_r_home() -> io::Result<PathBuf> {
 // Get the path to the R library
 fn get_r_library(r_home: &Path) -> PathBuf {
     let pkg_target_arch = env::var("CARGO_CFG_TARGET_ARCH").unwrap();
-    match (cfg!(windows), pkg_target_arch.as_str()) {
+    let r_library = match (cfg!(windows), pkg_target_arch.as_str()) {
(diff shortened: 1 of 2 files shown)


build.rs
```diff
@@ -194,13 +194,23 @@ fn get_r_home() -> io::Result<PathBuf> {
...
+    };
+
+    if let Ok(r_library_canonicalized) = r_library.canonicalize() {
+        r_library_canonicalized
+    } else {
+        // TODO: r_library might not exist in some types of installation that
+        // doesn't provide libR, R's shared library; in such a situation, just
+        // return this non-existent path for now. Probably the `rustc-link-search`
+        // setting itself is not used at all, but we are not sure at the moment.
+        r_library
```
Ilia-Kosenkov, line 213: Maybe we should not emit `rustc-link-search` if it does not exist at all? How about just returning `Option<PathBuf>`?

yutannihilation: Thanks, I moved back the `canonicalize()` to the bottom because

* canonicalization is only needed for rustc, and
* `.library` is also used in `<InstallationPaths>.get_r_binary()`, so this cannot be `None` on Windows.

Ilia-Kosenkov (approved):

PR #175, opened 2023-07-15
Update version to v0.6.0

yutannihilation:


yutannihilation:
/bindings

(In order to update the version header in the bindings)

CHANGELOG.md
eitsupi, line 1: Ending support for R < 4.2 seems like a breaking change.
So perhaps a version number of 0.6.0 would be more appropriate than 0.5.1?

yutannihilation: Oh, good point, thanks. I personally consider this change is trivial, considering you had been unable to use Rust on R < 4.2 on CRAN anyway because the CRAN machine for oldrel Windows didn't have Rust installed in the era when oldrel was R 4.1. But, I agree with you. Let's change the version to 0.6.0.

CGMossa (approved):

yutannihilation:
Thanks for approving! I'll publish this version.

PR #180, opened 2023-08-15
Added `NORET` fix for `bindgen`

CGMossa:


CGMossa:
With the introduction of `.enable_function_attribute_detection()`, the `NORET` issue should have been solved.
And thus one should have been able to remove it from `wrapper.h`. Locally, on Windows, this issue wasn't resolved. Now this PR is here to see if it has an effect at all on the bindings overall.

Thus, this is more for documentation rather than changing something..

PR #181, opened 2023-08-26
[experiment] Remove `OsStr` concerns

CGMossa:
Somehow this reduction in `build.rs` works and tests are fine.

Can you review the changes @yutannihilation ?

## Background

This was brought by Rust 1.72. It contained `OsStr` conversions. Plus a previous release contained similar conversion. So I thought maybe we don't need the OS-specific stuff. Turns out we might not need those at all.
**Note I'm not using Rust 1.72 specific features in this PR**.


yutannihilation:
> `OsStr` conversions

In my understanding, `OsStr` doesn't take care of the encoding differences.

CGMossa:
Alright, I went on an excursion

A few links:
<https://doc.rust-lang.org/std/process/struct.Output.html>

<https://doc.rust-lang.org/std/ffi/index.html#conversions>

<https://doc.rust-lang.org/std/os/unix/ffi/trait.OsStringExt.html>
<https://doc.rust-lang.org/std/os/windows/ffi/trait.OsStringExt.html>

I still do not understand how `Command::output` can yield a `u8`, even though it is encoded as a `u16` on something. Closing this for now..

yutannihilation:
Note for future reference: In the normal case, R probably outputs only ASCII characters, so we might get rid of Windows API. But, I'm not sure if there are any downsides and I don't feel this change is worth the risk.

CGMossa:
We can specify the encoding in the R cli. I don't know what that influences though.

I'd like to make the build.rs less complicated. I've got experiments to make it even more... Involved and wanted to have this attain minimal representation first.

yutannihilation:
> We can specify the encoding in the R cli

I'm thinking about the cases that happens outside of R, like "command not found" might be translated in each language or contain some Windows terminal-specific wide characters.

yutannihilation:
> I'd like to make the build.rs less complicated. 

To achieve simplicity, I think libR-sys should give up executing R at all and take all settings from envvars. This is also one of the problems in my mind in https://github.com/extendr/extendr/issues/583.

PR #182, opened 2023-10-04
Update bindings

yutannihilation:
This pull request is to reflect the removal of BLAS-releated functions in R-devel. This should have no problem.

[R News](https://cran.r-project.org/doc/manuals/r-devel/NEWS.html):

> Headers ‘R_ext/Applic.h’ and ‘R-ext/Linpack.h’ used to include ‘R_ext/BLAS.h’ although this was undocumented and unneeded by their documented entry points. They no longer do so.


CGMossa (approved):
I saw the news, and I think it should just be updated. I like the massive reduction on the bindings.. 

LGTM

yutannihilation:
Thanks!

PR #187, opened 2023-10-05
Merge extern blocks

CGMossa:
This reduces the bindings by 2k LOC.

- [ ] depends on #186


yutannihilation:
Looks good, but I'm curious why this requires the newer version of bindgen. It seems `merge_extern_blocks()` exists in more earlier versions. Do you find any problem?

CGMossa:
Oh! Right the dependence is because I was trying to show what PR to review first and approve first.

Because then you get a nice merging between the generated bindings.

yutannihilation:
Ah, I see, thanks. Then, I'm approving both now.

yutannihilation (approved):

PR #192, opened 2023-11-02
Update `bindgen` to `0.69.1`

CGMossa:
This should have absolutely no influence on bindings.

I saw a reaction at work, but that's because `llvm 17` apparently snaps more comments in the bindings.
Now that I'm at home, this difference is not present.

But then now we can upgrade this in peace.

Also added `-Wno-ignored-attributes` because when you run this with
```
cargo clippy --features use-bindgen -vv
```

you see a warning

```
[libR-sys 0.6.0] clang diag: C:/rtools43/x86_64-w64-mingw32.static.posix/include\stdio.h:169:18: warning: '__format__' attribute argument not supported: gnu_scanf [-Wignored-attributes]
```

which we don't really care about, as we are creating bindings and not compiling this.


Ilia-Kosenkov (approved):

CGMossa:
The call to `/bindings` doesn't work on this. I think GitHub CI fails here.

yutannihilation:
> I think GitHub CI fails here.

In case you don't notice, you can check the CI run here. 

https://github.com/extendr/libR-sys/actions

CGMossa:
Dear @yutannihilation, I'd love it if you could take a final look at this, and merge when if you find it done.

This was supposed to be a minor PR, but I knew there is always something coming up, and it is good these got addressed by you in #195.

These bindings also include the new `OBJSXP` value mentioned in https://github.com/extendr/extendr/issues/664.

yutannihilation (approved):
Looks good.

PR #193, opened 2023-11-02
Unify feature `runtime`

CGMossa:
Both `bindgen` and `clang` use the same feature, `runtime`, and it basically means the same thing, and fits in the same place. I just put them together.


bindings/bindings-macos-x86_64-R4.2.rs
```diff
@@ -1,8 +1,8 @@
 /* automatically generated by rust-bindgen 0.69.1 */
 
 /* libR-sys version: 0.6.0 */
-/* bindgen clang version: Homebrew clang version 17.0.3 */
-/* clang-rs version: Homebrew clang version 17.0.3 */
```
Ilia-Kosenkov, line 5: Why is it a downgrade?

CGMossa: I guess that's what was on the CI machines when I did this run :P What's really hilarious is that it affects what comments it recovers from source as well...

Ilia-Kosenkov (approved):

PR #196, opened 2023-11-12
Remove layout tests

CGMossa:
Remove these auto-generated layout tests. I don't think they have ever caught anything or helped with something. They could be plugged back in, if we refactor the generation of bindings with `xtask`-approach.


CGMossa:
This takes `/bindings` folder from 2.8 MB to 2.2MB.

build.rs
```diff
@@ -453,6 +453,7 @@ fn generate_bindings(r_paths: &InstallationPaths, version_info: &RVersionInfo) {
     // the resulting bindings.
     let mut bindgen_builder = bindgen::Builder::default()
         .emit_diagnostics()
+        .layout_tests(false)
```
Ilia-Kosenkov, line 456: Can we make it switchable? I definitely want to run layout tests when building libR-sys from scratch.

CGMossa: Could you explain more? The only time this has come up, was when bindgen was generating the layout tests wrong for some packed-something.
Is this for our CI benefit? For libR-sys developers?

CGMossa: Alright, I think I did it now! It just made this PR a little more complicated.

yutannihilation (approved):
Looks fine to me.

PR #197, opened 2023-11-12
Remove `BLAS.h`, `Linpack.h` and `Lapack.h` from bindings

CGMossa:
Inspired from a news item

> Headers ‘R_ext/Applic.h’ and ‘R-ext/Linpack.h’ used to include ‘R_ext/BLAS.h’ although this was undocumented and unneeded by their documented entry points. They no longer do so.

We actually don't use the wrappers to `BLAS.h`, `Lapack.h` or even `Linpack.h`, so we could remove them and improve clarity of bindings. If we were to include them, we'd actually need to write wrappers on top of them, as they aren't directly applicable. 

Diff on removing:
- [x] `BLAS.h` https://github.com/extendr/libR-sys/pull/197/commits/7a0b7ede0d559349d50341cd442364b90a7ad687 (-6K+10 LOC)
- [x] `Linpack.h` https://github.com/extendr/libR-sys/pull/197/commits/20d4f455d3748d964b7913e7e6b640ce812f6314 (no change)
- [x] `Lapack.h` https://github.com/extendr/libR-sys/pull/197/commits/dddf24ef9f5af07002269bcbcfc32a3d7babde6d (+4-88)


yutannihilation:
Just in case you don't notice.

* I believe libR-sys doesn't use `R-ext/Linpack.h`, so you don't need to exclude it.
* libR-sys does use `R_ext/Applic.h`, and the BLAS removal was already reflected by https://github.com/extendr/libR-sys/pull/182. But, this happens only on R-devel, and I have no strong opinion on removing them from the prior versions as well.

Also, I'm feeling this reasoning is not very true. These BLAS functions are independent from SEXP, so I don't think it's extendr's responsibility to provide such wrappers anyway.

> We actually don't use the wrappers to `BLAS.h`, `Lapack.h` or even `Linpack.h`, so we could remove them and improve clarity of bindings. If we were to include them, we'd actually need to write wrappers on top of them, as they aren't directly applicable.

Personally, I'd vote for removing `R_ext/Applic.h`.

CGMossa:
Thanks! As always, very good feedback! So I wanted to make a "Remove `Applic.h`" next, but I remembered why (apparently I added it), it is because it contains some routines that aren't Fortran once. Maybe #199 is a better solution to this?

CGMossa:
Closed in favour of #199

PR #199, opened 2023-11-12
Remove all Fortran items for bindings

CGMossa:
All Fortran things are renamed, so they have `_` appended to their name.
We can remove all of these with this PR.

We don't actually load `Rblas.dll` or `Rlapack.dll` (on Windows atleast), so I'm not even sure these things
would work, even if someone called them, meaning this isn't a breaking change.

This is an alternative to removing `Applic.h`, `BLAS.h`, `Linpack.h`, and `Lapack.h`, as this basically removes the Fortran in all of them, see #197.

Note that this PR is more or less another approach to #197. 
So only one should be merged; There are no real issues if both are accepted, though.


yutannihilation:
I can approve this because this doesn't remove what should not be removed, but I think I still don't understand the context. Why does Fortran matter here?

CGMossa:
I updated the description. These won't work anyways. We don't include the libraries in the compilation. Extendr shouldn't also provide bindings to blas or lapack anyways. I think Rcpp people use other libraries for linear algebra routines, because using these is not good.
I want these binding files to be nice, neat, and stuff like that. I'm constantly looking for ways to improve them.

And them having redundant, not linked, symbols is just wrong.

Plus.... It will make librsys smaller in size. With merge extern blocks we went from 4mb to 2.8mb, and this will further help.

CGMossa:
To ease your concern about excluding the wrong items, I've now made a regex, that matches only the items, with one single `_` at the end of the variable name. You can see in the diff, that now I don't exclude an item that was `__<NAME>__`. But this item is not useful anyways, so either approach works.

yutannihilation:
Thanks for the explanation.

> your concern about excluding the wrong items

I don't have such a concern. I just didn't see what's the criteria of what should be included or excluded.

yutannihilation (approved):

PR #201, opened 2023-11-12
Mark `SEXP` both `Send` and `Sync`

CGMossa:
While investigating https://github.com/extendr/extendr/issues/659 it became very clear, that atleast the unsafety associated with R's `SEXP` should be handled by `extendr` or any other library that piggybacks on these bindings.

Depends on

- [x] #196 
- [x] #199


CGMossa:
I don't agree with this PR anymore.

Issue #204, opened 2023-11-12
Reorganise GitHub Actions 


CGMossa:
I'd like to define workflows here that I see are present, so we can clean up the current CI we have going on.

For routine maintenance

1. Generate bindings from new versions of R, R-devel, etc.
2. `nonAPI.txt` also changes with these versions, and there is an action for it

For PR authors (contributors outside & inside the org)

3. Run tests, layout tests, on the generated bindings due to their changes, that they have not seen before, because they don't have all platforms available to them, nor all R versions installed.
4. Append their PRs with generated bindings, similar to (1), but should refer to current PR and push to current PR branch.

For maintainers / ambitious PR authors:

5. Run `extendr` main on PR'd generated bindings, I believe this is what is meant with #162

Ilia-Kosenkov:
Why separate 5 from a regular routine?

CGMossa:
I mean it is up for discussion.

I don't think every contribution warrants acceptance, based on its compatibility with current version of extendr. I don't think it is a burden to be put on outside contributors, for every little thing they PR. It might be up to maintainers / authors to benchmark this, or usher it in with upstream (to extendr) PRs, but not all contributions or contributors should be slammed with this.

There are thousands of improvements that could be done here, but we have interlocked the process between these two repos, I'd say, beyond reason.

yutannihilation:
IMHO, libR-sys is not the type of crate that has frequent updates except for regenerating the bindings against the latest version of R. So, I'm afraid optimizing it for pull requests is not worth. Frankly, I believe libR-sys should not be updated unless there's some strong reason.

For point 5, I guess the issue refers to tests on libR-sys, not on extendr. See https://github.com/extendr/libR-sys/pull/157.

CGMossa:
I disagree with this attitude in general. There are something wrong with the way github actions are setup at the moment. It blocks certain changes.

If libR-sys is only to serve extendr, then they should be merged. I believe if we had more users, there would be atleast monthly feature requests on this crate, and it is shockingly unfriendly to outsiders at the moment.

yutannihilation:
> I believe if we had more users, there would be atleast monthly feature requests on this crate

My take is the opposite. What's important in a sys crate is not "features," but stability. I use libR-sys outside of extendr, but never felt I should file a feature request because the current libR-sys is sufficient to me.

But, yeah, in actual, I think you are right. There are few libR-sys users like me. I think I'll create a simpler alternative for myself, so that's fine.

[...]

PR #205, opened 2023-11-16
Follow standard for opaque types in case of `SEXPREC`

CGMossa:
Since `SEXPREC` is an opaque type, that R doesn't want us to construct, or know anything about, or even
want us to make assumptions about. But, currently, we do! We assume that it is Send, Sync and Pin.
(see `cargo doc` pictures below).

I've made an issue about this https://github.com/rust-lang/rust-bindgen/issues/2685, because 
bindgen has an `opaque_type`, mechanism, and it is not producing a version with the 
`!Send,!Sync,!Pin`. But the https://stackoverflow.com/questions/38315383/whats-the-rust-idiom-to-define-a-field-pointing-to-a-c-opaque-pointer, and https://doc.rust-lang.org/nomicon/ffi.html#representing-opaque-structs, 
explains how to do this:

Going from

```rs
#[repr(C)]
#[derive(Debug, Copy, Clone)]
pub struct SEXPREC {
    _unused: [u8; 0],
}
```

to 

```rs
#[repr(C)]
pub struct SEXPREC {
    _data: [u8; 0],
    _marker:
        core::marker::PhantomData<(*mut u8, core::marker::PhantomPinned)>,
}
```

Results of this code can only be read-off `cargo doc`, so see the images below:

Currently:
image

Now:
image

Depends on:

- [ ] #208


CGMossa:
An example of the issues mentioned in #204 is here. What do I do now to showcase a PR? It breaks CI, so no-one will look. Even if the resulting bindings would work, if they were the ones saved and used in this PR.

CGMossa:
@yutannihilation

CGMossa:
@yutannihilation, since this works now, I'd like your thoughts on this.

yutannihilation:
Looks great, but let me confirm. In my understanding, `SEXPREC` is not a type of opaque type that needs to be exposed to the user because the API interface always uses `SEXP`, not `SEXPREC` directly. Does it make sense to block both `SEXP` and `SEXPREC` and manually define as below?

``` rust
pub type SEXP = *mut ::std::os::raw::c_void;
```

CGMossa:
I'm not sure. I don't think we should make `SEXP` a `void*`.

I believe Opaque Type atleast implies that it is a type we don't know how to construct, and that we are not allowed to construct. In Rust, that's accomplished with a private field, thus making it impossible to construct the type **outside of the defining module**.

I tried to do this:
image

But then `SEXP` becomes
image

Which is, in my opinion, the wrong definition. Because we can construct this.

I don't think we should blindly trust what `bindgen` produces. It is actively being developed.

I'm not sure if this PR has any impact. I think we should maybe include this PR, just for **correctness**.
I hope someone explains if this has an impact. Atleast this PR follows rust-documentation. Otherwise I don't know at all...

yutannihilation:
After reading [the Rustonomicon](https://doc.rust-lang.org/nomicon/ffi.html#representing-opaque-structs) again, I found my understanding about the opaque type was wrong, sorry. It seems you are right.

In this case, I'm still wondering if it's any better to expose the "type-safe" `SEXPREC` than to hide everything behind `c_void` because, unlike the Rustonomicon's example, the users always have to use `SEXP`. Anyway, except for this nitpicky point, I don't see any problem so I'm approving.

(JFYI, I decided to use `*mut c_void` in [my smaller alternative to libR-sys](https://github.com/yutannihilation/savvy/blob/703fa8c2532f523dd6ff8672482231b70e421f93/rlang-ffi-lite/src/lib.rs#L14), so I'll share if I find some difference that might matter)

yutannihilation (approved):

CGMossa:
Ok. Thanks for approval.
However, I'll wait until the fix for ci has been reviewed, and if someone on the bindgen part, or another rust ffi expert says something.

CGMossa:
This is the wrong approach. SEXPREC should be regarded as `c_void`. I'll put another PR, once I figure out how to make bindgen produce what I want..

PR #206, opened 2023-11-16
Make C-enums into idiomatic rust enums

CGMossa:
There are several "inferred" enums in the R-API. This setting ensures that
they become idiomatic Rust enums.
There are still missing ones, namely the macro-defined-enums, but maybe I'll add those somewhere else.

Issue #189, opened 2023-10-29
Generate bindings in xtasks, not in the build script


yutannihilation:
This is inspired by https://github.com/extendr/extendr/pull/611. I think this repository should also have xtasks.

There are mainly two motivations to separate the bindings generation from the compilation:

1. make the compilation of less troubles like https://github.com/extendr/extendr/issues/580
2. target wasm (I'm not sure if this is even possible, but I think it's a matter of time until not a few package authors and users wish the Rust-powered package on webR)

CGMossa:
I'm glad you like xtask! I've been working on a libR-sys xtask for a while now... But I was attempting to see, if I could make a new feature with it.

I'll post what I have later..

JosiahParry:
I would like to help work on this. What would we like the xtask ui to be like? 

If i understand correctly the xtask should be ran once per distro and R version. If a machine has more than one version of R available we can build bindings for all versions of R on that distro. We inherit the target from the machine running the xtask.

Should it look something like `libr-sys xtask --build --versions 4.2 4.3 4.4` ? Where we can provide multiple versions in the flag?

PR #208, opened 2023-11-17
Fix bindings failure on PRs

CGMossa:
This is not an implementation of #204.

Currently, PRs that change `libR-sys` cannot pass CI checks.

CI workflow right now: Runs checks + layout-tests. Uploads generated bindings (to other workflows). Then runs tests on precomputed bindings (that aren't "in" at the moment). This fails, because the generated bindings are not the precomputed bindings, the precomputed bindings are the ones from `main` / `master`.

Example of PRs that can't pass CI, even if the PR passes tests, are #206 and #205.

What this PR does is add a procedure that would commit the generated bindings, resulting from running CI, when the last commit message contains `[generate bindings]`.


.github/workflows/test.yml
```diff
@@ -226,14 +226,60 @@ jobs:
...
-        env:
-          RUST_TARGET: ${{ matrix.config.target }}
+      # # Run tests again using different bindings
+      # - name: Run tests on precomputed bindings shipped with libR-sys
+      #   if: matrix.config.skip-tests != 'true'
+      #   run: |
+      #     . ./ci-cargo.ps1
+      #     ci-cargo test -vv $(if ($env:RUST_TARGET -ne '') {"--target=$env:RUST_TARGET"} ) '--' --nocapture -ActionName "Running tests for target: $env:RUST_TARGET"
+      #   env:
+      #     RUST_TARGET: ${{ matrix.config.target }}
```
JosiahParry, line 236: Why are these commented out? Is it incase its easier to go backwards and uncomment it? Maybe its best to just delete?

CGMossa: I think we need to address this. This is the experience "everyone" else experience. Although this situation should be checked, in the first workflow, this is more like an extra check. But this fails. It fails because it doesn't use the newly generated bindings, because it can't.

.github/workflows/test.yml
```diff
@@ -226,14 +226,60 @@ jobs:
...
+  check_generate_bindings_flag:
+    name: Check if [generate bindings] is in latest commit message
+    runs-on: ubuntu-latest
+    outputs:
+      head_commit_message: ${{ steps.get_head_commit_message.outputs.HEAD_COMMIT_MESSAGE }}
+      generate_bindings: ${{ contains(steps.get_head_commit_message.outputs.HEAD_COMMIT_MESSAGE, '[generate bindings]') }}
```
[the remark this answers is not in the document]
CGMossa: This is what we have done all the time. We use GHA to generate all bindings, and then we commit those to the repo. We don't generate the bindings locally and upload them.
There is no build right, because we don't actually build something here, we merely generate the bindings in the `build.rs` script. In fact, I believe #189 is about moving the binding-generation to an `xtask` here.

CGMossa:
I have an issue now that it is running the workflow that pushes the bindings every time anything is run.
image

CGMossa:
This is fixed now, and I've rebased it to remove all the annoying commits.

.github/workflows/test.yml
```diff
@@ -226,14 +226,63 @@ jobs:
...
+    - uses: actions/download-artifact@v3
+
+    - name: Update bindings
+      run: |
+        # Update or add the bindings
+        cp generated_binding-*/*.rs bindings/
+
+        # Replace the default bindings
+        cd bindings
+        for x in linux-aarch64 linux-x86_64 macos-aarch64 macos-x86_64 windows-x86_64; do
```
Ilia-Kosenkov, line 273: Should this be arch-specific? I mean CI/CD runner os/arch specific?

MAINTAINERS_GUIDE.md
```diff
@@ -1,3 +1,17 @@
+# GUIDE
+
+## Commit newly generated bindings
+
+You last commit must contain `[generate bindings]`. You can make an empty
```
Ilia-Kosenkov, line 5:
```suggestion
Your last commit must contain `[generate bindings]`. You can make an empty
```

.github/workflows/test.yml
CGMossa: don't know. I copied what happens in the other task.
The thing is we might run more binding generations, than we want to keep. And they all will be dumped the same place. And this has to happen, otherwise there are no bindings to commit in the following workflow....

Ilia-Kosenkov: I mean, if this is a `linux-x86_64` runner, do we need to copy `windows-x86_64` bindings? At least I read it so here.

CGMossa: But it is from the previous workflows, which were done with the right platform. The download-artefact downloads what was uploaded in the previous workflows.

yutannihilation: > I mean, if this is a `linux-x86_64` runner, do we need to copy `windows-x86_64` bindings?

No. Committing the bindings updates should be done on a single runner. Otherwise there would be about 10 different commits for each platform and cause lots of conflicts.

> The thing is we might run more binding generations, than we want to keep

No. Thanks to `emit-bindings`, the binding generations happen only on the runners that you want to keep. This step is to update the default bindings for each platform.

.github/workflows/test.yml
```diff
@@ -226,14 +226,63 @@ jobs:
...
 
-      # Run tests again using different bindings
-      - name: Run tests on precomputed bindings shipped with libR-sys
-        if: matrix.config.skip-tests != 'true'
-        run: |
-          . ./ci-cargo.ps1
-          ci-cargo test -vv $(if ($env:RUST_TARGET -ne '') {"--target=$env:RUST_TARGET"} ) '--' --nocapture -ActionName "Running tests for target: $env:RUST_TARGET"
-        env:
-          RUST_TARGET: ${{ matrix.config.target }}
+      # # Run tests again using different bindings
```
Ilia-Kosenkov, line 229: Let's cleanup commented-out code.

Ilia-Kosenkov: @CGMossa I think this is the only missing piece

CGMossa: Done.

Ilia-Kosenkov (approved):

PR #215, opened 2024-01-07
Use `cargo:warning` to notify users during build time

Ilia-Kosenkov:
- Use `cargo:warning=Warning text` to properly notify users (regular `pirntln!()` do not really work)
- Put some dead code behind `cfg` (used only for bindgen)
image

### DIFF
--- a/build.rs
+++ b/build.rs
@@ -161,7 +161,7 @@ fn r_command<S: AsRef<OsStr>>(r_binary: S, script: &str) -> io::Result<OsString>
     // if there are any errors we print them out, helps with debugging
     if !out.stderr.is_empty() {
         println!(
-            "> {}",
+            "cargo:warning={}",
             byte_array_to_os_string(&out.stderr)
                 .as_os_str()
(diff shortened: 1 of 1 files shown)


build.rs
```diff
@@ -475,10 +481,6 @@ fn generate_bindings(r_paths: &InstallationPaths, version_info: &RVersionInfo) {
     let target_os = std::env::var("CARGO_CFG_TARGET_OS").unwrap();
     let target_arch = env::var("CARGO_CFG_TARGET_ARCH").unwrap();
 
-    println!(
-        "Generating bindings for target: {target}, os: {target_os}, architecture: {target_arch}"
-    );
-
```
CGMossa, line 481: i believe this is very helpful honestly, and some things should definitely be there in case we want people to run it with `-vv` flag to inspect it..

CGMossa (approved):
LGTM. Please address the comment I have before merging though.. Approving because why not..

CGMossa (approved):

PR #229, opened 2024-04-23
Cross-compile to intel mac from ARM mac

CGMossa:
Github's default macos-image has changed to ARM-based on. Therefore, we now have to cross-compile from ARM mac to intel mac.

Also, I believe the bindings are out of date, due to messy git rebasing done by me.

Note: This was a guided by @yutannihilation!

## Long explanation
I have been resolving conflicts in the opposite direction when doing `git rebase`. The term current, when rebasing,
refers to the main/master-branch, and not one's PR branch. Or maybe, I'm just mistaken how it should work. Regardless, I believe after a rebase, I should make a `[generate bindings]` commit, to truly ensure, that the bindings are correctly generated.


Ilia-Kosenkov (approved):
Since x86_64 diffs match, I believe it is safe. But let's monitor!

PR #230, opened 2024-04-24
Conversions between `bool` and `Rboolean`

CGMossa:
There are many places where `bool` and `Rboolean` are used and converted to and from each other in `extendr-api`.

This is an immediate convenience (safe) conversion that I think we should adopt.

There is a more controversial one, which is `From<u32>`, that I will propose after this one.


src/lib.rs
```diff
@@ -76,6 +76,24 @@ extern "C" {
...
 }
 
+impl From<Rboolean> for bool {
+    fn from(value: Rboolean) -> Self {
+        match value {
+            Rboolean::FALSE => false,
+            Rboolean::TRUE => true,
+        }
+    }
+}
```
Ilia-Kosenkov, line 86: Can you guarantee that there will not be any other case for `Rboolean` enum? From definition perspective it is exhaustive, but can `R` pass us some garbage to treat as `TRUE`?

CGMossa: R does pass us garbage to interpret as TRUE. I believe I wrote about it in the discord.. It is related to whether something is an S4 object. There, we get 16 and not 1.
But here, this is a conversion between rust's boolean and this representation for Rboolean. We don't have the u32-conversion here. If that conversion was present (which I may put in a PR going forward), then it would be `From<u32>`and it will have to interpret `>0` as `Rboolean::TRUE`.

CGMossa: And I can't cast a u32 to a Rboolean at the moment.

CGMossa:
CI is failing for arbitrary reason:
> Error: Failed to get R devel: Failed to get R 4.5.0: Failed to install R: Error: The process '/usr/bin/sudo' failed with exit code 100

src/lib.rs
CGMossa: @Ilia-Kosenkov what do you think?

Ilia-Kosenkov: What I meant is, R deals in `u32` (or `i32` or whatever). Then it has constants defined, like `FALSE` and `TRUE`. Based on that we generated an exhaustive enum. But if there will be a case that some function (due to binding generation) returns value other than `1` or `0` which is stored as `Rboolean`, then this will likely panic badly. If there are cases where we have functions in `libR-sys` that return `Rboolean` but can smuggle in `16` or any other value, we need to make this enum non-exhaustive and explicitly handle the case when it is neither `TRUE` or `FALSE`.

If however, this may never happen, then I am ok with current solution.

Ilia-Kosenkov:
```rust
  pub fn Rf_isS4_old(arg1: SEXP) -> i32;
```
Does it return `i32` or `Rboolean` in `libR-sys`?

Ilia-Kosenkov:
```rust
 pub fn Rf_isS4(arg1: SEXP) -> Rboolean;
```
This is your problem. If `Rboolean` is an enum of two values 
```rust
#[repr(u32)]
#[non_exhaustive]
#[derive(Debug, Copy, Clone, Hash, PartialEq, Eq)]
pub enum Rboolean {
    #[doc = ", MAYBE"]
    FALSE = 0,
    #[doc = ", MAYBE"]
    TRUE = 1,
}
```
but this function return `16u32`, Rust will likely explode with UB. You will write exhaustive `match` where non of branches (`TRUE` or `FALSE`) will work.

Ilia-Kosenkov: E.g. this https://github.com/rust-lang/rust-memory-model/issues/41

CGMossa: When I tried to make the example into a test, I got weird errors. The discriminant value is used in the hashing, and it is different, but I can't see it used in the partialeq, yet it still affect it.

CGMossa:
@Ilia-Kosenkov Honestly, I can't find a good way to eliminate this. I believe we should just redefine `Rf_isS4` with the correct conversion, and be done with it.

src/lib.rs
CGMossa: Please check what I've pushed now.. Otherwise, I'm calling it. I believe it would be a major disservice to not have `From<bool>`, especially since it is littered everywhere in the API.

We should just keep an eye on these things if they break. I already went through the remainder of the API, and only know of this one silly instance.

Ilia-Kosenkov (approved):
Yep, this looks solid!

CGMossa:
No-one merge please. I need to generate bindings first.

CGMossa:
Oldrel is 4.3 and so is release (4.3). R4.4 was released today. We must await the re-shuffling.

PR #225, opened 2024-04-22
Add Windows ARM bindings

CGMossa:
Fixes #190

PR #231, opened 2024-04-24
Finding all the rtools..

CGMossa:
* There is a new version of R released _right now_ so oldrel is now 4.3, at the same time, release is 4.3.
* Re-organised the rtools finding mechanism, because there are a lot of rtools now, see #225

There is Rtools for ARM, and Rtools for 4.4 that is added here.

I'd like to recap what is new here: Before, we specified the Rtools version manually for each new release, as we had to write which rtools version in the `matrix.config`. Instead now, we rely fully on setup-r and the version of Rtools it wants to use (they are tied to R versions now), and we greedily try to detect which rtools path is available.

*This means that the list has to be updated manually now*
This is still an improvement, because with Windows ARM, the path is a little different than the
pattern that we observed before.
_What we ought to do_ is either extract the rtools path from `PATH`, or scan for which path is present with R tools in it.

Neither R, not setup-r gives these rtools_soft paths, and thus we cannot exploit their configuration strategies.


.github/workflows/test.yml
```diff
@@ -135,24 +133,40 @@ jobs:
...
-          } elseif ($env:RTOOLS_VERSION -eq '43') {
-            $mingw_root = "C:\rtools43\x86_64-w64-mingw32.static.posix"
-          } elseif ($env:RTOOLS_VERSION -eq '42') {
-            $mingw_root = "C:\rtools42\x86_64-w64-mingw32.static.posix"
+          # Source: https://github.com/r-lib/actions/blob/b7e68d63e51bdf225997973e2add36d551f60f02/setup-r/lib/installer.js#L471
+          $directories = @(
+            "C:\rtools44-aarch64\aarch64-w64-mingw32.static.posix",
+            "C:\rtools44\x86_64-w64-mingw32.static.posix",
+            "C:\rtools43\x86_64-w64-mingw32.static.posix",
+            "C:\rtools42\x86_64-w64-mingw32.static.posix",
```
Ilia-Kosenkov, line 144: How about maybe scanning for `rtools??` folders? Otherwise, you will have to update this if there will be `rtools45` and so on.

CGMossa: We already do that. And the ARM support will be tricky.
I also don't know how to write this with little failure points.

Ilia-Kosenkov (approved):

PR #234, opened 2024-04-30
Minor missing instruction

CGMossa:
This little instruction was missing from what can be c/p if needed.

Very minute, minor addition


CGMossa:
It has come to my attention, that we may not necessarily need _both_ of those files for things to work. Therefore, it might be, that the instructions have to be revised. Regardless.

PR #235, opened 2024-05-03
Modify bindings to generate an emscripten target

JosiahParry:
The PR adds emscripten-wasm32 bindings **enabling extendr to work with webr**. 
This PR modifies the yml to also generate emscripten bindings. We have tested and compiled `rsgeo` to wasm using this approach.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -335,6 +335,12 @@ jobs:
           # Choose the newest version except for devel
           ln --force -s "$(ls -1 ./bindings-${x}-*.rs | grep -v devel | sort | tail -1)" ./bindings-${x}.rs
         done
+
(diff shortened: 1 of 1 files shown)


CGMossa (approved):
LGTM: One comment is, I thought the `build.rs` had to modified somehow. That it should be modified because otherwise it would actually _not_ generate the bindings for emscripten, if it was asked to do so.

But what this does, is it provide a compatible cached emscripten bindings, so when using libR-sys without `use-bindgen` feature, it will use that to link.

So there is still an ambiguity here that we need to address.

CGMossa:
Let us wait a little on merging this. Josiah and I have to do some diligence.

JosiahParry:
Agreed. I'll move into draft.

PR #237, opened 2024-05-09
feature:  `non-api` adds non-API items to bindings

CGMossa:
This adds the ability to generate bindings that includes non-API items.

There are several reasons for this

- **Developer debugging** As R-core is marking more items as non-API,
it might be desirable to have a feature to allow for experimenting.

- It is only CRAN, and only a "NOTE": You might have good reasons for including aforementioned non-API items, therefore, these may be argued for.

- `libR-sys` should not solely be aiming to target CRAN, extendr (extendr-api) crates.. and more or less aim to serve as a sys-crate to R API in general. 

Note that these bindings are not cached, which means, to include non-API items, your system
must be setup to generate these bindings somehow.

Therefore, to generate these bindings you must enable this feature in the `Cargo.toml`, e.g.

```toml
libR-sys = { version = "*", features = ["non-api", "use-bindgen"] }
```


CGMossa:
I've rebased, which broke this feature. Then I fixed it, and I've added the `non-api` feature to the test API.

This is now ready for review. Let me know what you think.

Ilia-Kosenkov (approved):

PR #238, opened 2024-05-09
Remove `clang-sys` dependency

CGMossa:
This PR accomplishes three very distinct things:

- ~Adds a missing `:` infront of `cargo::` everywhere.~
- ~Adds `metadata`, which have been missing from `build.rs` forever~
Update: the above items are unnecessary. rust 1.77 have a new _syntax_ for `build.rs`. I didn't
catch that initially. This has now been reverted, and MSRV for this PR is the same.
- Removes the inclusion of `clang`/`clang-sys`, by instead of the elaborate processing, we instead, ignore all items coming from headers that aren't the R-headers.


yutannihilation:
I didn't know the new syntax. Is it your intention to drop 
the support on Rust <1.77?

> Stick to `cargo:` only if the support of Rust version older than 1.77 is required.

https://doc.rust-lang.org/cargo/reference/build-scripts.html

CGMossa:
Wow! Thanks! I did not know about syntax change in cargo 1.77. Yesterday, while I was trying to make this strategy work, I tried everything, and then I thought we had used the wrong syntax.

I've reverted this.

@Ilia-Kosenkov You asked about the `Rcomplex` thing. Since we are filtering elements in a new way, maybe now,
there is an element not being filtered, or something like that. But I just checked, and our definition of `Rcomplex` is still the same.
The items from `wrapper.h` are being considered just fine.

CGMossa:
So one explanation to the extra new items that somewhat looks like:
```c
enum {SORTED_DECR_NA_1ST = -2,
      SORTED_DECR = -1,
      UNKNOWN_SORTEDNESS = INT_MIN, /*INT_MIN is NA_INTEGER! */
      SORTED_INCR = 1,
      SORTED_INCR_NA_1ST = 2,
      KNOWN_UNSORTED = 0};
```
these are items that stem from an _anonymous_ C-`enum`. Before, with the clang-processing, we would just ignore it.
This time, it is being included. There is zero harm with it being present.

bindings/bindings-linux-aarch64-R4.3.rs
```diff
@@ -271,6 +270,12 @@ pub enum Rboolean {
     #[doc = ", MAYBE"]
     TRUE = 1,
 }
+#[repr(C)]
+#[derive(Debug, Copy, Clone)]
+pub struct Rcomplex__bindgen_ty_1 {
```
Ilia-Kosenkov, line 275: I am still confused why is this type here? It does not seem to be used any where else.

CGMossa: I don't know. But it was replaced. It is a non issue for now. Bindgen is really also more of a magic than science, and as long as it doesn't produce errors, I'm fine with it.

Ilia-Kosenkov: Should we put it in the black list then?

CGMossa: I didn't think that was possible. But it is! Will commit in a second.

Ilia-Kosenkov (approved):

PR #244, opened 2024-06-13
Emit bindings for `oldrel-1`

### LINKED ISSUE #243: Duplicate definition of Rf_isS4
I'm trying to complite a package on a machine that seems to require rebuilding the R library.

When I try to compile `libR-sys` I get the following compilation error.

```
   31    Compiling libR-sys v0.7.0 (/home/andres/libR-sys)
   30 error[E0428]: the name `R_altrep_Coerce_method_t` is defined multiple times
   29    --> /home/andres/libR-sys/target/debug/build/libR-sys-12a2d53e43b7ed60/out/bindings.rs:578:1
   28     |
   27 578 | / pub type R_altrep_Coerce_method_t =
   26 579 | |     ::std::option::Option<unsafe extern "C" fn(arg1: SEXP, arg2: ::std::os::raw::c_int) -> SEXP>;
   25     | |_________________________________________________________________________________________________^ `R_altrep_Coerce_method_t` redefined here
   24     |
   23    ::: src/lib.rs:80:1
   22     |
   21 80  | / pub type R_altrep_Coerce_method_t =
   20 81  | |     ::std::option::Option<unsafe extern "C" fn(arg1: SEXP, arg2: SEXPTYPE) -> SEXP>;
   19     | |____________________________________________________________________________________- previous definition of the type `R_altrep_Coerce_method_t` here
   18     |
   17     = note: `R_altrep_Coerce_method_t` must be defined only once in the type namespace of this module
   16
   15 error[E0428]: the name `Rf_isS4` is defined multiple times
   14     --> /home/andres/libR-sys/target/debug/build/libR-sys-12a2d53e43b7ed60/out/bindings.rs:1698:5
   13      |
   12 1698 |     pub fn Rf_isS4(arg1: SEXP) -> Rboolean;
   11      |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ `Rf_isS4` redefined here
   10      |
    9     ::: src/lib.rs:83:1
    8      |
    7 83   | pub unsafe fn Rf_isS4(arg1: SEXP) -> Rboolean {
    6      | --------------------------------------------- previous definition of the value `Rf_isS4` here
    5      |
    4      = note: `Rf_isS4` must be defined only once in the value namespace of this module
    3
    2 For more information about this error, try `rustc --explain E0428`.
    1 error: could not compile `libR-sys` (lib) due to 2 previous errors
```

I think the error is pretty self explanatory, there is definition duplication between `lib.rs` and the different bindings. 

I am 100% willing to help out with this issue. However, some guidance on the implications of changing these values would be appriciated as guidance.

andyquinterom:
Fixes #243 

I want to add cases for version that do not fit on the `devel`, `oldrel`, etc schema. 

Many companies still use R 4.2 for example so I believe this is relevant. 

@CGMossa Perhaps adding an `emit-binding` could also be helpful?


CGMossa:
Yes you need to add lines to the matrix, where emit binding is true.

Then you need a commit with the text containing [generate bindings] for the ci to push generated commits onto this PR.

If you do this properly, I can approve it and it will be done much faster I guess...

andyquinterom:
@CGMossa 

Done I guess. Let me know if there is any left

CGMossa:
Let's run ci... And see

andyquinterom:
Once CI is done I'll send a new commit adding MacOs x86 to the matrix.

andyquinterom:
@CGMossa the `Check if [generate bindings] is ` is failing. Perhaps because I have by branch on a fork?

CGMossa:
I would also like to take this opportunity, to tell you, about some of the constraints we need to consider here. If we want `libR-sys` + `extendr` to be CRAN complaint, there is a limit on how large a package we can send. For every version of R, and every supported platform, we have a relatively large text file embedded in these packages, and they have a noticeable effect on the package size. 

Are you sure your company can't upgrade to 4.3?

andyquinterom:
> I would also like to take this opportunity, to tell you, about some of the constraints we need to consider here. If we want `libR-sys` + `extendr` to be CRAN complaint, there is a limit on how large a package we can send. For every version of R, and every supported platform, we have a relatively large text file embedded in these packages, and they have a noticeable effect on the package size.
> 
> Are you sure your company can't upgrade to 4.3?

It's not mostly on my side. It's specifically for the package `orbweaver`. We are in the process of getting it to our users, but many of them are on different R versions.

If we want R packages (with Rust) that are widely used I think having wide support for atleast recent R versions is important.

A strategy could be compressing the bindings into a tar ball with gzip? This is already done with the vendored solution for CRAN. This could be integrated in the `build.rs` step maybe?

CGMossa:
> @CGMossa the `Check if [generate bindings] is ` is failing. Perhaps because I have by branch on a fork?

Apparently, this is an issue, but I fixed it via `ref: ${{ github.event.pull_request.head.sha }}`. Thank you ChatGPT 🤷

andyquinterom:
> > @CGMossa the `Check if [generate bindings] is ` is failing. Perhaps because I have by branch on a fork?
> 
> Apparently, this is an issue, but I fixed it via `ref: ${{ github.event.pull_request.head.sha }}`. Thank you ChatGPT 🤷

Amazing! I will push a commit filling in the blanks now that everything seems to work

andyquinterom:
@CGMossa The update bindings commit does not seem to have updated the r 4.2 bindings. I'll investigate

andyquinterom:
The job meant to create the PR with the generated bindings failed because of the same `ref` to `sha` error.

https://github.com/extendr/libR-sys/actions/runs/9495244689/job/26167470383

I just sent a commit to fix it

CGMossa:
> The job meant to create the PR with the generated bindings failed because of the same `ref` to `sha` error.
> 
> https://github.com/extendr/libR-sys/actions/runs/9495244689/job/26167470383
> 
> I just sent a commit to fix it

I don't think that's the cause. We have two mechanisms for generating bindings in PRs. I just told you about one. And you can see that it worked, before the PR has now updated bindings for all versions. Except for R 4.2. I've looked at the generated file that comes out of the CI run you specified, and it looks like it has the right R version, and that it has changed. But somehow it is not being pushed.

.github/workflows/test.yml
```diff
@@ -269,7 +273,7 @@ jobs:
     steps:
     - uses: actions/checkout@v4
       with:
-        ref: ${{ github.event.pull_request.head.ref }}
+        ref: ${{ github.event.pull_request.head.sha }}
     - uses: actions/download-artifact@v4
```
CGMossa, line 277: I think this needs to be changed back to `ref`. It worked fine before the `sha`.

andyquinterom: I'll try reverting. Seems weird since it did not work before

CGMossa:
Sorry, I've looked as thoroughly as I can right now, and I don't see what the issue is. We made a lot of progress on this.. I hope you can nail the last issue, but I personally just can't see it. Feel free to ask questions, I'll try to answer as fast as I can.

andyquinterom:
@CGMossa I think the issue is that I am on a fork. Which makes sense why CI cannot find a branch, since this branch does not exist on the main repository. Any chance we could take this branch to the repo?

CGMossa:
Let me test this.. Because clearly some files were updated just now. Look at the file menu, and see the changes. Those come from the update mechanism. After this, I'll push your branch to this repo and see if that's the issue.

andyquinterom:
As you can see on flow https://github.com/extendr/libR-sys/actions/runs/9495808517

It seems /bindings is running on commit 09d76ad which is the master branch not this branch.

This is the reason the new bindings are not being included.

andyquinterom:
I think I narrowed it down to 

```
      - name: Switch branch (/bindings command)
        if: github.event_name == 'issue_comment'
        uses: r-lib/actions/pr-fetch@v2
        with:
          repo-token: ${{ secrets.GITHUB_TOKEN }}
``` 

I still believe it may be an issue due to forks

andyquinterom:
@CGMossa what do you think?

CGMossa:
I can't get it to work.

PR #242, opened 2024-06-09
Full R API bindings

CGMossa:
# WORK IN PROGRESS

PR #245, opened 2024-06-13
POC: Tar ball compressed bindings

andyquinterom:
This is just an idea I had to make the bundle size of libR-sys smaller for CRAN compliance.


CGMossa:
Okay, while this is _nice_ I believe we do something like that "downstream" in `rextendr` when we want to vendor packages. Or atleast some of our users do. I think this effort should be put there instead. https://github.com/extendr/rextendr

I would like to say that I've benefitted from seeing this approach. It looks so clean...

You seem very capable at this stuff, would you mind taken a look at my little POC over at #242? It has been a long time ambition of mine, and I think I've gotten it far, but Windows and Paths are still messing with me.

.github/workflows/test.yml
```diff
@@ -339,9 +342,10 @@ jobs:
         cd ..
 
         # detect changes (the code is derived from https://stackoverflow.com/a/3879077)
-        git add bindings/
+        tar cJfh bindings.tar.xz bindings/
+        git add bindings.tar.xz
         git update-index --refresh
-        if ! git diff-index --quiet HEAD -- bindings/; then
+        if ! git diff-index --quiet HEAD -- bindings.tar.xz; then
```
CGMossa, line 348: This should still be a diff on the _text files_ and then just push the tar-file if needed.

I also think the tar-file should be pushed using `git lfs` right?

PR #247, opened 2024-06-17
Lock a dependency to ensure MSRV of `1.67.1`

CGMossa:
Ran `cargo update -p home@0.5.9 --precise 0.5.5` and then the MSRV is 1.67.1.

We are commiting a lockfile here to ensure MSRV is always checked for CRAN complaint.


CGMossa:
See https://github.com/rust-lang/rust-bindgen/issues/2850 for the source of this advice.

PR #250, opened 2024-06-28
Detach `bindgen` for next release

CGMossa:
For various reasons, we have to make a new release of `libR-sys`.

This PR removes all uses of `bindgen`, and removes all features associated with generating bindings.
The currently available bindings in `/bindings` are those that we will have until #189 is addressed.

After this PR, the MSRV for `libR-sys` is at:  1.58.1.


JosiahParry:
Thank you! I think we need move the deleted build contents into `./create-bindings.rs` and in the `Cargo.toml` write `exclude = ["create-bindings.rs"]`

CGMossa:
> Thank you! I think we need move the deleted build contents into `./create-bindings.rs` and in the `Cargo.toml` write `exclude = ["create-bindings.rs"]`

I would vote for keeping this PR clean. It removes code, and we publish it, without any clutter. Then after that, we'll make a release, and then we can bring those bits back in.

JosiahParry:
I get that. Okie. Let's merge this so we can unblock CRAN fixes.

JosiahParry (approved):
Approved. Let's merge and work on the xtask

Issue #251, opened 2024-06-29
Generate non-API bindings via xtask 


JosiahParry:
We need to remove bindgen from the dependencies of libR-sys to ensure that the MSRV is met for CRAN (so silly that an organization that doesn't use Rust can specify an MSRV because they don't want to run `rustup update`). 

Additionally, since R 4.5 is formalizing the C API there are a lot of things that are causing warnings. These two things together make for a bit of a headache. 

For those who want to use non-API (in case anyone is using an environment iterator or something), we need to make those available to them.

This has to be done via xtask and stored in existing bindings a la #189

PR #253, opened 2024-07-01
Manual update of non-API

CGMossa:
R-devel is actively redefining what is, and what is not public API.

This PR applies the restrictions found here https://github.com/extendr/libR-sys/pull/252#issuecomment-2200273018.
Note that we are following @yutannihilation's advice here on `PRESEEN` and `SYMVALUE`. 

We broke the CI for bindings generation, so a manual update of the binding is necessary. See #251.


JosiahParry (approved):
Apologies for the delay here. Thank you @CGMossa !!!! 💕

PR #254, opened 2024-07-02
bindings: Remove remaining `Rf_isS4` and `R_altrep_Coerce_method_t` as we manually define those

CGMossa:
Somehow, these bindings was not removed from the cached bindings. I suspect our way of detecting `devel` might have been wonky, or a git merge went awry. 

Either case, here's a potential hotfix.


JosiahParry (approved):

CGMossa:
I don't think so. Their removal was working on the other generated bindings files.

PR #255, opened 2024-09-09
Stop updating bindings using CI

yutannihilation:
After https://github.com/extendr/libR-sys/pull/250, the scheduled execution of CI keep failing because it no longer emits bindings.

```
cp: cannot stat 'generated_binding-*/*.rs': No such file or directory
```
(https://github.com/extendr/libR-sys/actions/runs/10753234750/job/29822527827)

Whenever it fails (i.e. once per 3 days), GitHub emails me about the failure. I really don't know why GitHub blames me even after the HEAD commit of `master` is not mine. I know this is no one's fault, but this is quite annoying. So, I'd like to stop this by deleting the schedule. I think you can restore these lines when you get the generation of the bindings ready.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,9 +12,6 @@ on:
   issue_comment:
     types:
       - created
-  # Run this once per three days
-  schedule:
-    - cron: '29 17 */3 * *'
   # This can also manually run
   workflow_dispatch: {}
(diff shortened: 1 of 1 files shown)


CGMossa (approved):

yutannihilation:
Thanks!

PR #258, opened 2025-02-27
Fix version finding on multiple lines

### LINKED ISSUE #257: Weird error while deteching R verison
There is this weird bug (that I cannot reproduce outside of a very specific environment) when restoring a package from `renv`.

Basically, `build.rs` is unable to get the R version due to a `NULL` being printed before the version. There is no global R profile. The package install ok with `remotes` (even when renv is initialized). 

Image

I think we could fix this by running this in a while loop or using an iteraton while a skip while.

Image

andyquinterom:
* Closes #257


andyquinterom:
After this change #257 's example works

image

